PR #171, opened 2015-06-02
Minimum SDK 15

ivanarrizabalaga:
- Goes back to SDK 15
 - .idea files are gitignored (they are automatically reloaded from build.gradle so there is no need to control them)
 - A couple of tests are updated since some buttons have changed
 - Removes an unnecesary 'onClick' in **dashboard_row.xml**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,8 @@
-# malariapp iml
-malariapp.iml
+# iml files (idea)
+*.iml
+
+#Idea custom configuration
+.idea/
 
 # Built application files
 *.apk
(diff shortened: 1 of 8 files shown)


ivanarrizabalaga:
@ifoche

Pull request for minimum SDK + some minor stuff.

By the way, I noticed that scores are calculated in a different manner now I got an '8' where I used to get a '4' while executing the tests. Is that ok?

ifoche:
Ivan, I don't know if you're still using bugs_v7 for anything else, can you delete that branch if you're not using it anymore?

ivanarrizabalaga:
I'm still using it to deal with the rest of assigned bugs.

ifoche:
ok, perfect. It was just to say unused branches have to be deleted. If you're using it, is ok.

PR #207, opened 2015-06-11
Dashboard list alignment

ivanarrizabalaga:


ifoche:
I have two questions here Ivan:
1 - As long as I can remember, these padding setting modifications where there to avoid unnecessary blank lines between different surveys. Have you checked that removing them doesn't waste screen space?
2 - Why is that lorem ipsum for? If it is needed, ok, but this should be in every language to avoid problems with apk generation
Can you check both?

ifoche:
As I think today you're out for the weekend, I'll do it now. I'll leave here the comments saying whatever I do

ifoche:
Ok, I've just checked it. Apparently, no wasted space between surveys when they collapse, so nothing done in this sense. For the 2nd, I've just added the string to the other languages, but please, could you next week explain here what is that string for? (just to be aware). I proceed to accept the pull request

ivanarrizabalaga:
Hi @ifoche!
I'm about take the car to Madrid so I just can give you a quick reply.

Custom views such as TextCard run its init method while previewing an xml
layout.

Because of this our TextCard was causing a crash in preview so I asked for
'isInEditMode' in order to distinguish both scenarios.

The lorem ipsum text is just a helper that shows that mock text while
previewing so you can figure out its final look just previwing the layout
(was useful for me at least).

Hope it helps!

ifoche:
Thanks for the explanation. It has a lot of sense, and will help a lot having the android previews repaired for being able to design the layouts. Thank you very much.

Issue #127, opened 2015-05-19
Review gradle lint errors


ivanarrizabalaga:
Right now:

./gradlew lint 

shows 13 errors + 219 warnings.

Most of them are not important, but we need to clean them up sooner or later.

ivanarrizabalaga:
Errors
---------------
 - Correctness
    - [ ] 13	Error MissingPrefix: Missing Android XML namespace  
 - Correctness:Messages
    - [ ] 1	Error StringFormatInvalid: Invalid format string    

Warnings
---------------
 - Correctness

    - [ ] 1	Warning InconsistentArrays: Inconsistencies in array element counts
    - [ ] 2	Warning AppCompatMethod: Using Wrong AppCompat Method
    - [ ] 1	Warning DuplicateIncludedIds: Duplicate ids across layouts combined with include tags
    - [ ] 1	Warning InlinedApi: Using inlined constants on older versions
    - [ ] 1	Warning OldTargetApi: Target SDK attribute is not targeting latest version
    - [ ] 1	Warning SimpleDateFormat: Implied locale in date format
    - [ ] 27	Warning UnusedAttribute: Attribute unused on older versions
    - [ ] 1	Warning InflateParams: Layout Inflation without a Parent
    - [ ] 1	Warning GradleDependency: Obsolete Gradle Dependency
    - [ ] 2	Warning GradleOverrides: Value overridden by Gradle build script
    - [ ] 1	Warning Deprecated: Using deprecated resources
    - [ ] 8	Warning PxUsage: Using 'px' dimension

- Performance

    - [ ] 12	Warning ObsoleteLayoutParam: Obsolete layout params
    - [ ] 7	Warning ViewHolder: View Holder Candidates
    - [ ] 1	Warning UseValueOf: Should use valueOf instead of new
    - [ ] 15	Warning DisableBaselineAlignment: Missing baselineAligned attribute
    - [ ] 1	Warning InefficientWeight: Inefficient layout weight
    - [ ] 3	Warning Overdraw: Overdraw: Painting regions more than once
    - [ ] 44	Warning UnusedResources: Unused resources
    - [ ] 22	Warning UselessParent: Useless parent layout

- Usability:Icons

    - [ ] 1	Warning IconColors: Icon colors do not follow the recommended visual style
    - [ ] 1	Warning IconMissingDensityFolder: Missing density folder

- Usability

    - [ ] 1	Warning TextFields: Missing inputType or hint

- Accessibility

[...]

PR #218, opened 2015-06-16
Fix some xml lint errors

ivanarrizabalaga:
Solves #127


ifoche:
One question Ivan, is it strictly needed to have in each element of the iqatab.xml the namespace declared? Is is not enough to declare it once in the top element (LinearLayout) that contains the rest?

ivanarrizabalaga:
I think it is only required at the top element (that's exactly the change I
made, removing that declaration from every TableRow).


[image: http://]
Ivan Arrizabalaga
[image: http://]about.me/ivan-arrizabalaga
<http://about.me/ivan-arrizabalaga>

ifoche:
Oops, sorry, I interpreted the changes upside-down :).

PR #253, opened 2015-07-06
Jose is pushing

josemp10:
The push (Zambia) is coming!!! And Custom Tabs strikes back!


ifoche:
After the merge, compositive scores are failing. Jose is going to give a look to it, restore them and after that, apparently, push would be prepared to be integrated into development

PR #290, opened 2015-07-13
Adds progress between tab changes

ivanarrizabalaga:
@ifoche PR done, i think that the loading progress per tab works ok and but personally I dont like the final result (as a user) very much. 

It looks like the survey is less responsive, but I have not tested it all together (last optimizations + progress per tab).


ifoche:
Thanks, let me check it

ifoche:
We're keeping this PR until Friday and there, if we're not able to reduce composite score tab latency, then we would introduce this progress bar for that tab

ifoche:
I've merged hotfixes_to_release_9 to this loading_tabs branch, as it looks like we may need it. The only problem that it has is that the new asynctask, compared to the old showTab(), does not adapt his views when customize fonts is selected. Can you give a look to this @ivanarrizabalaga ?

ivanarrizabalaga:
@ifoche Sure, I going to check it now

josemp10:
Sorry guys, if you both think that these progress bars are ugly when they
are being rendered in the screen, I think we should not include this in the
next release (9C, 9D... whatever, I can not remember anymore). I mean, we
should show this result to Rodolfo first and let him to decide... (another
reason to change to DBFlow)

ifoche:
Jose, the progress bar was something Rodolfo asked. We were not including this because we had the hope of being able to easily reduce significantly the latency when changing tab, but sadly, as @ivanarrizabalaga has discovered it is not possible in the short term, I'm afraid we won't have any other chance at the moment.

PR #308, opened 2015-07-22
percentage completion bug solved

ifoche:
After testing it, it looks like we're changing to "Prepared to send" state when we're almost at 100% . But that bug will be treated in a different issue


ifoche:
That issue is #309

PR #325, opened 2015-08-08
Build variants

ifoche:


ifoche:
I'm going to merge this, but uploaded databases are Zambia ones, so further work will be needed in order to set the proper DBs to each country

josemp10:
I have checked the csv files and I think there is a bit of
misunderstanding here, so let me try to clarify this:

Eds and hnqis are going to be slightly different apps (also they are
pushing to different servers), and , as far as I know, this was the main
reason to use the build variants in the app.

EDS are the countries: zambia, Malawi, Mozambique, Tanzania, and Ghana

HNQIS are just for psi programs : SF and ICM

It is not possible to have sf and icm under EDS, or Malawi, Tanzania,....
under HNQIS. The reason is just because they are for different clients and
also the scope is different.

Please let me know if it is unclear, because this is quite important to
understand.

FYI, We are not expecting to have other new surveys in the next months.

Also, you can remove the headercustom and questioncustom csv files as they
are no longer needed (all the logic for custom tabs is included in the
question and header csv files).

adrianq:
@josemp10 
- At the moment we have edsGhana, edsMalawi, edsMozambique, edsTanzania, edsZambia and hnqis (which contains SF & ICM). This has been like this since we introduce the build variants. What is the problem? Is not exactly the same you have said? Please, have a look at the project and let us know if you can see any different between what you says and what we have implemented.
-I have checked the header and question custom csv files. I can only find them in Zambia and as far as I know you were the one in charge of updating the db and implementing the custom tabs. Maybe you updated it in the drive folder but you forgot to update it in the code. Anyway it was not being used in the populate db. I will delete these files. Btw, the csv in the drive  was different from the one in the repo. Maybe did you forget as well about updating the csv contents? We have done it for you but if you can check that the content is correct. Hopefully, as we have centralised the csv in the repo you won't have this kind of forgets anymore.

josemp10:
Ey Adrian... I have seen an edsSFICM folder, so that was the reason of my
concern. Glad to know that it was well understood.

Custom tabs are updated with all the surveys in Google drive (custom
questions and headers csv files are not being used since several months
ago). We agreed to have them all in g. Drive and not in the code, so I
guess I did not forget anything in that regard.

PR #390, opened 2015-09-06
Child-parent relation to match tables

josemp10:


ifoche:
We should keep track of which DBs have already their parent-child relationship included in the new relationship tables. I assume that, for the moment, we have this only in HNQIS variant. Maybe a shared doc? It would be fine not to rely just on our memory for that.

ifoche:
@ivanarrizabalaga this impacts you. Development branch has from now on the parent-child relationships included using the new schema for HNQIS variant.

PR #391, opened 2015-09-06
Synchronize development with last R10 hotfixes

ifoche:
After R10 was published, some bugs where found and this branch has received a few hotfixes. In order to keep development updated with these changes, we merge master to development. There are conflicts due to a doubled change that is exactly the same in both branches, so conflict resolution shouldn't be very difficult


ifoche:
With this changes incorporated into development, @ivanarrizabalaga , I think you can continue introducing parent-child relationship using the new schema.

PR #397, opened 2015-09-09
Bug with comp. scores and custom tabs solved!

josemp10:
Phantom Questions!!


ifoche:
Composite scores were not being calculated for custom tabs. In custom tabs, parent questions were phantom questions in some cases (just to group them) and this was interfering with the composite scores calculation (the old algorithm didn't detect those questions). @josemp10 solved it

Issue #392, opened 2015-09-07
Settings location is optional


ivanarrizabalaga:

ivanarrizabalaga:
@adrianq I need credentials for the DHIS server to try the new location settings. (Dont write them here since this is public)

Issue #346, opened 2015-08-14
Refactor Parent-Child relation into new relation db schema


adrianq:

ivanarrizabalaga:
Queries required for navigation:

/*Get parent question*/
select q.*
from question q 
left join questionoption qo on q.id_question=qo.id_question
left join match m on qo.id_match=m.id_match
left join questionrelation qr on m.id_question_relation=qr.id_question_relation
where qr.id_question=228 and qr.operation=1;

/*Get parent question + option*/
select qo.id_question,qo.id_option
from questionoption qo
left join match m on qo.id_match=m.id_match
left join questionrelation qr on m.id_question_relation=qr.id_question_relation
where qr.id_question=228 and qr.operation=1;

/*Get children for question + option*/
select q.*
from question q left join questionrelation qr on q.id_question=qr.id_question
left join match m on qr.id_question_relation=m.id_question_relation
left join questionoption qo on m.id_match=qo.id_match
where qo.id_question=227 and qo.id_option=130 and qr.operation=1;

ivanarrizabalaga:
Enable dynamically output for sql queries in DBFlow:

com.raizlabs.android.dbflow.config.FlowLog.setMinimumLoggingLevel(FlowLog.Level.V);

ivanarrizabalaga:
Counting the current number of optional questions that should be answered according the values of the parent questions:

select q.*
from question as q 
left join questionrelation as qr on q.id_question=qr.id_question
left join answer as a on q.id_answer=a.id_answer
left join match as m on qr.id_question_relation=m.id_question_relation
left join questionoption as qo on m.id_match=qo.id_match
left join value as v on qo.id_question=v.id_question and qo.id_option=v.id_option
left join option as o on v.id_option=o.id_option
where a.output<>7 and v.id_survey=1 and qr.operation=1

PR #401, opened 2015-09-10
Parent child

ivanarrizabalaga:
#346, #392


ifoche:
@ivanarrizabalaga I don't know if I'm doing something wrong but it seems to me that this branch doesn't work. I've incorporated to it the last changes made to development, but if you go back to your last commit, you'll see that it didn't work anyway before those changes. I mean, I can identify 2 main problems:
- The parent-child relationships are not being detected (you can compare with master branch to check how it should look like)
- The custom tabs are crashing
Am I wrong?

ivanarrizabalaga:
@ifoche, I did it under the hnqis variant I didnt test it for any other
variant (maybe the data is not prepared for each of them).

Anyway I will take look to see buts happening.

ifoche:
You're absolutely right regarding custom tabs. For this release we don't have to adapt the code that is not HNQIS. My mistake. Please don't waste your time with that.

However, regarding parent-child relationships not being detected in EDS...although it is not a goal for this release, I cannot see what's so specific in HNQIS that cause EDS parent-child relationships not being shown. I think this could be revealing a problem we were not aware about.

I've just checked and going back to your last commit, HNQIS was showing its parent-child relationships, so I have introduced the bug when merging. I'll give a look to this and update you soon.

ifoche:
@adrianq gave me the solution to EDS parent-child relationships. It was also my mistake. In the DB there are no parent-child relationships for the rest of the variants (and they weren't supposed to), sorry. Don't waste a minute with this, I'll solve the bug I introduced.

ivanarrizabalaga:
@ifoche great news!

Issue #406, opened 2015-09-17
Dashboard modifications for R11


ifoche:

ifoche:
![dashboard_modifications](https://cloud.githubusercontent.com/assets/10931645/9936841/3ae29276-5d5e-11e5-918a-74c7b7baca21.png)

ifoche:
This cover 3 main issues:
- [ ] Overall Composite Scores (on Dashboard and Feedback Activity)
- [ ] New Score column in sent assessments
- [ ] Traffic light bullet

PR #408, opened 2015-09-18
Feedback sent

ivanarrizabalaga:
Includes customviews in feedback and #406.

@ifoche It looks like the radio button 'onlyIfFailed' is not scaling properly and remains the same size no matter what preference you got.

I'd rather do the PR anyway since the rest of code looks fine, and probably you know how to fix it quicky.


ifoche:
ok, thanks @ivanarrizabalaga . I'll take care of it.

PR #431, opened 2015-09-23
Pick location on create survey

### LINKED ISSUE #426: Change GPS logic to be recovered when survey is created, not when survey is sent

ivanarrizabalaga:
Closes #426

### DIFF
--- a/app/src/main/java/org/eyeseetea/malariacare/CreateSurveyActivity.java
+++ b/app/src/main/java/org/eyeseetea/malariacare/CreateSurveyActivity.java
@@ -20,6 +20,10 @@
 package org.eyeseetea.malariacare;
 
 import android.app.AlertDialog;
+import android.content.Context;
+import android.location.Location;
+import android.location.LocationListener;
+import android.location.LocationManager;
 import android.os.Bundle;
 import android.util.Log;
 import android.view.LayoutInflater;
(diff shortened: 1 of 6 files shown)


ifoche:
little typo in the name of the branch :)

ifoche:
@ivanarrizabalaga my only concern here is about the onLocationChange event. Does it means that, if the user accesses the survey being away from the health facility, the location of this survey is going to change? I mean, is this a callback that will be updated as long as the survey is active?

ifoche:
@ivanarrizabalaga I've just tried to push some surveys, doing the following tests:
- Create a survey in a normal speed, without GPS activated, and push
- Create a survey waiting a little bit in the create survey activity, with the GPS activated, and after that push
- Create a survey with WiFi activated but not GPS, waiting a little bit in the create survey activity, and after that push

After these tests I've accessed the DHIS server to see the data. I can see the rest of the data element, but I can't see in any case the GPS location. Am I doing something wrong to test it? Have you checked it in the remote server?

josemp10:
@ifoche,

a) Are the GPS coordinates (0,0) in DHIS2?
or
b) can not you see the GPS coordinates data points?

If b) maybe there is a problem with the configuration...

josemp10:
@ifoche, I dont know if you are receiving this. There was a problem with
the coordinates of HNQIS in DHIS2 (configuration problem). Now, it should
be solved (you may need to clean the cache to see the changes). HNQIS was
not capturing GPS coordinates, so all of them may be 0. In other words, you
need to test it again.

@ivanarrizabalaga does not have a user on the server. Will create one for
him, so he will be able to test it by himself.

ifoche:
@josemp10 ok, thank you for fixing that. I'll test it again. @ivanarrizabalaga didn't have a user, and thank you for creating one, but in the meanwhile he had ours for testing.

ifoche:
@josemp10 I've tested it again and now I'm not even able to see the whole tree of orgUnits at the left in DHIS2. May my credentials be different now? However, I can see the orgunit that I'm uploading and there there is no GPS related field at all (even after pushing a survey)

josemp10:
Try to open chrome in incognito mode... Let me know

ifoche:
That made the trick!

ifoche:
Good, then we can validate the development, thank you @ivanarrizabalaga for the work, and thank you @josemp10 for fixing the server bug

ivanarrizabalaga:
Uhmmmmm It looks like there was a lot of action while I was sleeping, :).

Thank you guys, so finally is it working @ifoche??

ifoche:
:) yes, it is merged and working. My question yesterday was whether or not, you had checked it on the server, because I couldn't see them. To summarize, Jose found that the server was not prepared to receive the GPS coordinates (from DHIS side) and when he included it I was able to see them. So yes, it looks like it is done :)

Issue #422, opened 2015-09-21
HTML parser is not interpreting everything


ifoche:

ifoche:
Feedback is being fed by HTML tagged text. In general, everything is being formatted properly but some things are not being properly read:
- [ ] Bullet lists (<ul><li/></ul>) are not being read
- [ ] Images appear as a blue square. We have to guess if this is because it is not reaching the URL, but however, we need to render images (even if we are force to download them and have them in local)

ifoche:
[this](http://stackoverflow.com/questions/11214001/show-ul-li-in-android-textview) explains that our parser is not reading ul-li tags. We need to change the parser to allow it

ifoche:
[here](http://stackoverflow.com/questions/7424512/android-html-imagegetter-as-asynctask) you'll find another interesting link

ivanarrizabalaga:
A pseudo approach:

 - Rename URLImageParser into CustomParser
 - Implements Html.ImageGetter and  TagHandler

In the method:

**public void handleTag(boolean opening, String tag,
                                 Editable output, XMLReader xmlReader);**

check for the tags "ul", "li" and follow a sample to turn it into the proper string.

Check this link http://stackoverflow.com/questions/11214001/show-ul-li-in-android-textview

ifoche:
done by using [Kutsi code](https://bitbucket.org/Kuitsi/android-textview-html-list)

PR #432, opened 2015-09-24
Full html supported in feedback

### LINKED ISSUE #418: Introduce images in feedback content
we need to check if this is working already for valid local and remote URLs

ivanarrizabalaga:
Closes #418, #422


ivanarrizabalaga:
@ifoche finally works fine with the ImageGetter, super cool! :D

ifoche:
Here, as we talked yesterday, there would be only missing the ul-li tags, that are not being interpreted yet. I will keep that task and reassign to me, to do it this weekend

PR #433, opened 2015-09-25
Feedback gui changes

### LINKED ISSUE #421: Feedback GUI changes

ivanarrizabalaga:
Closes #421 and #413


ifoche:
Just one thing before approving the PR. I've seen the way you changed the checkbox text size has been by setting a 18sp fix size to it. That will work for one specific screen size and font size configuration, but will fail in the rest of the cases as long as I can think. So I'll approve the PR, but I'll introduce a new task to solve this.

PR #436, opened 2015-09-26
Newdatabase

josemp10:


ifoche:
@josemp10 just one consideration before merging this. This is ICM and there are 2 facilities called "KE - HNQIS SF pilot test facility 1" and "KE - HNQIS SF pilot test facility 2". Is that ok? Aren't they supposed to be for SF and not ICM?

PR #446, opened 2015-09-28
percentage now being properly calculated

ifoche:


ifoche:
Solution has been a little ugly, so I've opened a new issue to clean it (#447) for release 1.2

PR #454, opened 2015-09-28
Calculates main score in prefeedback, calculated spinner read only

### LINKED ISSUE #349: Polish Match Feature
- [x] Disabled Spinner

ivanarrizabalaga:
Closes #349 and #440.


ifoche:
I'm testing this and I can't see this working neither on 4.4.2, nor on 4.2, 5.0... But it looks strange. Is not that this is not disabling the spinner, it is that even match doesn't work at all. I've tested on virtual an real devices. Am I supposed to have something in the DB that you didn't uploaded yet? (something like a new kind of answer output...) (working in HNQIS kenya variant, just in case you want to check it out)

ivanarrizabalaga:
Match is working ok for variant **hnqisCongoICM**, I guess there must be something wrong with Kenya (data)

ifoche:
ok, thank you. I'll create an issue for @josemp10 to check this and merge PR now

ivanarrizabalaga:
Example (Kenya, Questions.csv):

Question 24 --> Answer 14 with Output 1 (should be 10)

PR #537, opened 2015-11-03
Remove unnecesary uses-permissions

idelcano:


ifoche:
Hi @idelcano, I think these permissions were already removed from this branch, so I cannot accept the PR. Can you check it out again in visitors_try branch (do a git pull before)? Just to confirm. Thanks

idelcano:
It works.
Only we need remember one thing.
if we can use Debug class.
we need:

Android profiling using Debug class
App must have these permissions
<uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE" /> 
<uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />

And remove:

    <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" tools:node="remove" />

ifoche:
I think I'm not properly understanding what you want to say. Your commit is introducing tools:node=remove, that is a way to resolve conflicts when merging Android manifests. I suppose that is the solution to avoid Google Play introducing its own permissions, so in case of conflict, those permissions wont be added. Ok, I'll approve that, but this line (that is in the visitors_try branch already)
      <uses-permission android:name="android.permission.READ_CONTACTS" /> 
wont it be redundant with the line you added?
     <uses-permission android:name="android.permission.READ_CONTACTS" tools:node="remove" /> 

I would say that first line must be removed, don't you think? can you do it and push it to the branch so I can merge the rest?

idelcano:
It is totally true. Sorry.
I checked all the uses-permission independient of pictureapp.

Uses-permissions merged by Gradle:
GET_ACCOUNTS; 
USE_CREDENTIALS; 
READ_EXTERNAL_STORAGE; 
WRITE_EXTERNAL_STORAGE;
I removed all.
And i removed read_profile and Read_contacts.

I tested in a real device 5.0.1 and api 23,21,17  and works.

PR #538, opened 2015-11-03
License style

idelcano:


ifoche:
Hi @idelcano . Here in malariapp there is also the Portuguese as an eligible language, can you update also Portuguese licence? Thanks!

idelcano:
I dont have the text of the license in portuguese:
http://www.gnu.org/licenses/gpl-3.0.html
The website is available in portugese-Brazilian but the license is in english for that language.

ifoche:
I've searched quickly without any result either, so I'll merge this and we'll leave the license in English for Portuguese people. Thank you

Issue #544, opened 2015-11-08
Start using the real user in the app


ifoche:

ifoche:
We were creating our own "new User()" in our database and not filling it with the info available from the remote server. Now that we're pulling the data, we can translate User table from SDK User pulled.

ifoche:
As we talked before, this information comes from UserAccount table

PR #551, opened 2015-11-10
Pull working after relogin, logout async completed

### LINKED ISSUE #540: pull-not-finishing bug
The pull works the first time (after login).

After that a pull under demand looks like loads all the stuff but somehow the data taken from the server is made of nulls.

If you logout an try to login again the pull doesnt crash the app but it looks like it is not completely finished.

ivanarrizabalaga:
Closes #540

### DIFF
--- a/app/src/main/java/org/eyeseetea/malariacare/BaseActivity.java
+++ b/app/src/main/java/org/eyeseetea/malariacare/BaseActivity.java
@@ -36,10 +36,14 @@ import android.view.View;
 import android.view.Window;
 import android.widget.TextView;
 
+import com.squareup.otto.Subscribe;
+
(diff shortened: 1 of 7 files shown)


ifoche:
Hi @ivanarrizabalaga. As User is not being created anymore after login, we're leaving the user table empty, and the user is needed in order to create surveys (is one its FKs). It is what #544 try to describe (I tell you this because this is the PR that is removing our user creation). So I'm accepting this PR, but please, it would be good to implement the UserAccount(SDK) to User(EST) translation as next task. Thank you very much!

ivanarrizabalaga:
@ifoche I know, that is my main task today. ;)

ifoche:
Thanks!

PR #580, opened 2015-11-19
Dashboard show last survey for orgunit

idelcano:


ifoche:
As I already told you. The only thing that last for checking is to verify that no one is using the session adapter to get the surveys apart from monitoring section (that you already managed to give it the correct list of surveys).

Issue #561, opened 2015-11-12
Push before delete unsent surveys


ifoche:

ifoche:
There is already another issue that will talk about this task (see )

ifoche:
When the user press on pull option (that must be renamed to "Metadata refresh"), if there are unsent surveys, the user must be asked whether or not (s)he wants to push them before deleting all the data.

ifoche:
@ivanarrizabalaga, as @idelcano has been very quick with his assigned issues, I'm reassigning this task to him. Please forget about it

idelcano:
I finished this task, but i need try in actual version when we had a surveys to be sent

PR #586, opened 2015-11-20
Push before delete unsent surveys

idelcano:


ifoche:
This closes #561

ifoche:
sorry @idelcano . Please keep this branch in local, but I have finally decided not to accept this PR. Instead of that, I've uploaded the branch push_before_pull that I had in my local repo with the final solution to the problem. I'll do a PR now, so please, if you find there any bug you think you had properly managed in your solution, let me know.

PR #590, opened 2015-11-22
Screen switches off crash

### LINKED ISSUE #585: When pull is in a real device and the screen switches off, an exception is thrown

idelcano:
@ifoche 
Added pulling control with preferenceshared (if you don´t like i remove it).
Fixed the progresactivity dialog as models dialog.
closes #585

### DIFF
--- a/app/src/main/java/org/eyeseetea/malariacare/EyeSeeTeaApplication.java
+++ b/app/src/main/java/org/eyeseetea/malariacare/EyeSeeTeaApplication.java
@@ -21,6 +21,8 @@ package org.eyeseetea.malariacare;
 
 import android.app.Activity;
 import android.content.Context;
+import android.content.SharedPreferences;
+import android.preference.PreferenceManager;
 import android.support.multidex.MultiDex;
 
 import com.crashlytics.android.Crashlytics;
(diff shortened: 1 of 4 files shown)


ifoche:
@idelcano sorry about asking this but, what is the point of making a difference between the first pull and the rest of the pulls? I mean, why are we storing "incompleted_pull", "first_pull" and "completed_pull" and not only a boolean variable "pull_in_progress" with "true" or "false"? Sorry if it is too clear, I didn't meditate it too long

idelcano:
True and true. It should be called pull, no first pull.
And true, i will remove it and use boolean value.
I dont use it becouse  i cant  save  boolean  in a String type
(getstring/putstring) obviusly ... i forget i can save boolean type in
shared preferences.
Totally true. Sorry!

idelcano:
it is fixed.

I think it makes more sense to use "pull_metadata" that "pull_in_progress"
Because when we launching the application,the value "pull_in_progress" should be true and this is not true, however, if "pull_metadata" is false, it is true....

I fixed using pull_metadata variable name.
Anyway, if you want i use pull_in_progress.

idelcano:
I fixed other problem.. I think is related.. and it could be in this branch.

If you press outside of the dialog, the activity dont do nothing and the dialog disappeared.
I make the dialog as model dialog.(only add setcancelable(false)).

ifoche:
Thank you very much! The code looks very nice.

PR #620, opened 2015-12-02
New MZ Org Units

josemp10:


ifoche:
@josemp10 Does this DB update need to generate release? do we have to wait for more changes before doing it? Thank you

josemp10:
@ifoche, sorry, I lost my connection after the PR... So, yes, there are
some new facilities in DHIS2 to incorporate to the Mozambique APK (EDS -
R10). Would it be possible to generate a new APK by tomorrow at any time
(R10)? As far as we know, there are no more changes for EDS coming.

Thanks!

ifoche:
ok @josemp10, perfect, thank you for the info. I think it is possible.

PR #651, opened 2015-12-11
Matching sdk event dates with surveys dates

### LINKED ISSUE #627: Dates refactor

ivanarrizabalaga:
Closes #627.


ifoche:
@josemp10 We're starting to use "dueDate" column from the Event table, that was apparently unused in the server, for registering the schedule date. Can you confirm that there's no problem at all using that field? Do you better prefer us not to upload that date, and just keep it only calculated offline?

josemp10:
@ifoche, yes I think you can use the "dueDate" for that porpuses (I can not
think in a reason for not using that field), and it is probably a good idea.

ifoche:
ok, thanks @josemp10 . @ivanarrizabalaga , wouldn't this need a DB BaseMigration derivative class to be able to upgrade the DB without deleting the app? Can you add it to the PR?

PR #655, opened 2015-12-14
Question type

### LINKED ISSUE #235: Create a new table to render the answers
Drop Down List? Radio Buttons Vertical? Radio Buttons Horizontal?

ivanarrizabalaga:
Closes #235


ifoche:
As we've just talked, here also the Migration would be missing

PR #656, opened 2015-12-16
Assess tab

### LINKED ISSUE #636: Assess Tab

idelcano:
closes #636

### DIFF
--- a/app/src/main/java/org/eyeseetea/malariacare/DashboardActivity.java
+++ b/app/src/main/java/org/eyeseetea/malariacare/DashboardActivity.java
@@ -58,11 +58,16 @@ public class DashboardActivity extends BaseActivity {
     DashboardUnsentFragment unsentFragment;
     DashboardSentFragment sentFragment;
     LocalActivityManager mlam;
+    static boolean viewFeedback;
 
     @Override
     protected void onCreate(Bundle savedInstanceState) {
         Log.d(TAG, "onCreate");
         super.onCreate(savedInstanceState);
+        if(viewFeedback) {
+            viewFeedback=false;
+            finishAndGo(FeedbackActivity.class);
+        }
         setContentView(R.layout.tab_dashboard);
         try {
             initDataIfRequired();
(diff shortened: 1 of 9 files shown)


ifoche:
@idelcano some comments:
- Could you change the way you're deciding where to go next in ProgressActivity.java? At this moment you're deciding it depending on the content of the msg variable. If you look it carefully you'll realize that this comes from getDoneMessage() and there, that message decided from calling isAPush() and hasAPullAfterPush() methods. Maybe, instead of comparing that string, you can call yourself those functions and decide it depending on that.
- Try not to name i to the Intent object, because i is not intuitive. If you call intent, at least the reader will know that this is an intent. Even better if you call it something like currentIntent or callingIntent...something that distinguish that intent from any other that may exist in the same function...if that's the case.
- You're using a class var to determine if you need to go to ask for feedback after the push, if I'm not badly interpreting your code. I think that could be better done by adding the var to the intent, don't you think? I mean, instead of doing
DashboardActivity.viewFeedback=true;
finishAndGo(DashboardActivity.class);
maybe you can add a parameter to finishAndGo to add vars to the intent, and pass this var with the intent

idelcano:
1. True... Sorry. But, i need show the feedback button only if i have one push.
. if hasAPullAfterPush() is false, and isAPush is true, how can be sure it is only one push?. At this time, the application only send more than one push only if you refresh metadata, but, is not best choose when you want show the feedback button?

2. True too, but i copy that from other place in our code( i take a look and change in the other place too). https://github.com/EyeSeeTea/malariapp/search?utf8=%E2%9C%93&q=i.getIntExtra

3. It was my fist option. I was trying to do that.. but it cant work, i´m not sure why, i´m feeling lost time, (I think the dashboard was reloaded from service in other place and my intent from progressactivity was null) i will take a look again.

ifoche:
@idelcano : 
1. I think there's no difference from what I said. You're deciding depending on a variable that is set using those functions. If you have that problem using the functions, you'll have that problem with your solution too. If you want, we can quickly meet now and give a look to this. Please let me know

2. Yes, probably that piece of code must be refactored too, then. Can you do it and add it to the PR?

3. Ok, then please create an issue for that refactor and assign it to me, I'll take care of that asap.

idelcano:
@ifoche I think its all done. Thanks for your time !

ifoche:
@idelcano did you tested it in all the situations (Pull / Push & Pull after Push)? (Just to know).

idelcano:
No! Sorry  i m not test

ifoche:
Ok. Can you do it? Thanks

idelcano:
Yes, its tested now.
The pull is so fast... :)

PR #657, opened 2015-12-17
Improve tab

### LINKED ISSUE #637: Improve Tab

idelcano:
At this moment, we can look the unsent completed surveys with score in the dashboardSentFragment.
closes #637

### DIFF
--- a/app/src/main/java/org/eyeseetea/malariacare/DashboardActivity.java
+++ b/app/src/main/java/org/eyeseetea/malariacare/DashboardActivity.java
@@ -32,24 +32,18 @@ import android.util.Log;
 import android.view.Menu;
 import android.view.MenuItem;
 import android.view.View;
-import android.widget.AdapterView;
-import android.widget.Spinner;
 import android.widget.TabHost;
 import android.widget.TextView;
 
-import com.raizlabs.android.dbflow.sql.language.Select;
 import com.squareup.otto.Subscribe;
 
-import org.eyeseetea.malariacare.database.model.OrgUnit;
-import org.eyeseetea.malariacare.database.model.Program;
+
 import org.eyeseetea.malariacare.database.model.Survey;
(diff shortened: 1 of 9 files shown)


ifoche:
Hi @idelcano , just one thing before merging it. Could you move the "new Select()" calls to the DB Models and call them from there? I know that we have lot of "Select" calls spreaded through the code, but we have to stop doing that, as DB interactions should be limited to the DB models. I will afford that refactor soon

idelcano:
@ifoche  it is ready.

ifoche:
@idelcano I'm having a bug when sorting the improve results by the orgUnit column. It is like it were ordering only from the second row on. If I press the OrgUnit column I would expect to have all the surveys that belongs to the health facility 1 together, and those belonging to health facility 2 together, but instead of that, I'm getting the first row HF1, and after that every HF2, and then the rest of the HF1...Can you replicate this bug? Can you solve it before I merge this branch?

idelcano:
yes @ifoche i will take a look.!

ifoche:
@idelcano I'm merging development into this branch, as it had some conflicts. I'll tell you in a few minutes but, can you then pull the changes and try again with them? I recommend you to stash your changes before pulling mine

idelcano:
ok!

2015-12-21 12:22 GMT+01:00 Ignacio Foche <notifications@github.com>:

ifoche:
Finally it was not a few minutes. It is taking so long, as I've encountered some strange problems when merging development into your branch. I'll take a look again in a couple of hours, but don't wait for me to upload the merge before. Sorry...but I'm not sure where the problem comes from

idelcano:
Ok! 
I fixed the bug and pushed the branch. 
What problems do you have? maybe i know where look

I moved the Select() query to the db model.
i didnt read your comment before... sorry.

ifoche:
ok, it's done and working, thanks @idelcano

PR #664, opened 2015-12-22
Planning tab

### LINKED ISSUE #639: Planning Tab

ivanarrizabalaga:
Closes #639


ifoche:
@ivanarrizabalaga a few comments to the PR before accepting it:
- Can you keep the strings.xml default server modification (developer server) local in your computer?
- Can you move the "new Select()" calls involved in your modifications to the POJOs? I mean, for example, in line 136 of CreateSurveyActivity.java move it to ProgramExtended POJO as a static function to be able to call it like ProgramExtended.getAllPrograms(). We have a lot of those type of things but I think that if each one of us refactor each "new Select" call that we see in every modification we do, we will have soon every DB call in the POJOs. There are more than one (other example is line 296 of PlannedItemBuilder)
- Can you add in lucid chart the new Program <--> OrgUnit relationship, the new Survey Status and the new SurveySchedule table?
- Typo in line 270 of PlannedItemBuilder.java (Buids instead of Builds)
- Can you tag with a TODO the mocked productivity retrieval in PlannedSurvey.java so we can easily recognize it to change it whenever productivity is in server side?
- Can you move the function "findScheduledDateBySurvey()" to Survey POJO?
- In planning_header_row.xml , can you write and use dimen.xml dimensions instead of directly hardcoding them there? This is something general for every xml you edited, I've seen it more than once.
- In planning_schedule_dialog.xml can you use our CustomButton instead of a normal Button? I understand that this is to be able to click on the text and get the CalendarPicker, but it is preferable to add the listener to a normal CustomTextView or add customize the CustomButton instead of using a normal Button, as otherwise that won't change it's font size with the preference. 
- In general, in xmls, you're setting the typeface property. That should be managed using the CustomViews capabilities (app:tFontName="@string/medium_font_name"). Can you change it?

In terms of functionality:
- When you create a new assessment from the shortcut in "Future" assessments, and you go back to the dashboard, you find that assessment in the "Never assessed" row. I think that is a wrong behavior. That assessment should stay on "Future" assessments, don't you think? Maybe the problem comes from not having a scheduled date yet when we create a new assessment (so processAsNever() function recognized it as a Never Assessed survey).

Thank you in advance for it!

PR #670, opened 2015-12-30
Contextual action menu branch

### LINKED ISSUE #661: Contextual action menus

idelcano:
closes #661


ifoche:
error-string-missing

@idelcano looks like there's some string missing, don't you think? Can you please check it to see if you have uploaded all the strings?

ifoche:
@idelcano I think I've found a bug here (update the branch to test it because I've merged development into it). If you try to create a new survey (I used KE - HNQIS SF pilot test facility 1 as OrgUnit and the HIV program) and mark it as completed, then immediately it tries to send it (that's good) but if it finds an error in the server side (it happened to me) then it starts an infinite loop trying to resend it continuously. In my opinion it should wait some time before retrying. Can you take a look to this?

idelcano:
Yes @ifoche , but maybe we can think more conditions to cancel the push (in
pictureapp the push is cancelled if the user not have access to the org
unit, or the org unit is banned).
The error of the serverside only exists in the dev server.
I informed to @adrianq , and i commented it in other issue.
(anyway, i will do it, if the push fails, the tries will stop)

2016-01-04 10:05 GMT+01:00 Ignacio Foche <notifications@github.com>:

ifoche:
Yes @idelcano, I think it would be better to cancel the push after a certain number of failures (moving the survey to something like a FAILURE state) but I think that is something we have to discuss with the client. Please don't make the app to stop trying to send a survey after a failure (because then, we would need to add a mechanism to retry). Just make the tries to be separated like 30 seconds in case of failure. I know that the failure is something that happens in the server, but maybe we can detect that failure taking into account that if you are trying to resend a survey that you previously tried to send, then it must have failed that push process. Can you do that?

idelcano:
It is done. You can increment the time of the failed push in the string: PUSH_FAILED_PERIOD https://github.com/EyeSeeTea/malariapp/pull/670/files#diff-01dafb3fd0217441330103cfc8300094R299
It is now delayed between 25-30 seconds.

ifoche:
ok, thank you very much

Issue #689, opened 2016-01-13
Change graphs for assess tab


ifoche:

ifoche:
- [ ] Unsent fragment
- [ ] Create survey fragment
- [ ] Survey fragment

PR #700, opened 2016-01-18
Surveytabbuttons next previous

### LINKED ISSUE #698: Add tab buttons (previous/next) in Surveyfragment
Add tab buttons (previous/next) in Surveyfragment

idelcano:
Closes #698 related : #689


ifoche:
@idelcano is there any special reason for having nextButton and previousButton as class variables instead of limiting them to just the method where they're used? If there isn't, can you change it?

idelcano:
True! Done.

2016-01-20 4:05 GMT+01:00 Ignacio Foche <notifications@github.com>:

PR #706, opened 2016-01-21
Persist eventsuid in db

### LINKED ISSUE #701: Persist Events UIDs in the DB

idelcano:
closes #701


ifoche:
@idelcano did you tried to push with this code? I'm getting a 404 ERROR message

idelcano:
@ifoche it´s true, sorry, but it is solved now(and i re-check in DB the event creation date was saved).

ifoche:
Hi again @idelcano . I've just seen that here we would be missing one thing here. When we pull the metadata and data, we're not setting the events id in their respective surveys (in our DB), I mean, in the visitors part (https://github.com/EyeSeeTea/malariapp/blob/persist_eventsuid_in_db/app/src/main/java/org/eyeseetea/malariacare/database/iomodules/dhis/importer/ConvertFromSDKVisitor.java#L304-304). Can you add this so we would have in our DB all the event ids of the previously sent surveys?

Thank you very much

idelcano:
It is done.

ifoche:
great! thank you very much

Issue #690, opened 2016-01-13
Change graphs for improve tab


ifoche:

ifoche:
- [ ] Sent fragment 
- [ ] Feedback fragment

PR #707, opened 2016-01-21
Change improve tab

idelcano:
Closses #690


ifoche:
Hi @idelcano,

this has a bug, that has something to do with the commit 50c2563350fa6a2797e8900fb5745a7f943da952 that is touching the way backgrounds are calculated. The thing is that first line OrgUnit is hidden while it shouldn't. However, that commit doesn't say in its subject that it is touching the algorithm to show/hide the orgUnit title, so can you revert it and add new separate commits to a) redo the rest of the things you were doing in the commit b) touch that algorithm but with careful to avoid introducing that bug separately?

I attach you a screenshot to let you see the bug
orgunit_bug

idelcano:
@ifoche I did little refactor to do more clean the code of setbackgroundcolor(in the mentioned commit)I identified the bug. 
I will create a new issue. 
The bug is in the repaint of the rows.
It is appearing if the screen is more long than the list.
I think the problem is we set the color/OrgUnitTitle in base at position(and rows before the repainted row), and the position is changed if we move the list.

PR #713, opened 2016-01-22
Make fragment list load on background

### LINKED ISSUE #704: Make all lists loads in background when there is a fragment transition

idelcano:
closes #704
The next step is refactor reloadDashboard from this branch.


ifoche:
Just to keep trace of it, I've changed the constants that defined the filter titles into internationalized strings called from the same place. Otherwise we were not going to have translation of those strings

PR #742, opened 2016-01-29
Fix actionbartitle and change tab unselect icon background to transparent

### LINKED ISSUE #719: Change Actionbar titles and Tab button
Change the actionbartitles (show max 3 lines), and change the unselect tab buttons for a button with transparent background.

idelcano:
Closes #719


ifoche:
@idelcano have you tested it? It's strange but is like the title is not properly coordinated with the tab you change to. i.e. if you move to a tab different than improve, you get the name of the previous tab you visited instead of the tab where you are. Please see the attached images to understand better what I'm saying (in the first one, I access monitor from assess tab, in the second one, I went back to planning from monitor):

bug_actionbar_1

bug_actionbar_2

idelcano:
Sorry @ifoche, it is true (I only paid attention in the style).
It is solved and pushed.

PR #752, opened 2016-02-09
Load org unit levels

ivanarrizabalaga:
Closes sdk#18


ifoche:
Sorry @ivanarrizabalaga . I think there's a problem here. Somehow, the commits from the push tests have finished also in this branch. Was it something we talked about? If there's no special reason for doing it, I would prefer to merge this branch just with the code for orgUnit levels (a revert is ok). 

Thanks in advance

ivanarrizabalaga:
@ifoche, I will try to fix asap.

ivanarrizabalaga:
@ifoche, sorry it took me a couple of days to find the right time to revert this but now it is finally done, ;)

PR #763, opened 2016-02-11
Feedback changes

### LINKED ISSUE #759: Feedback changes

idelcano:
closes #759
I revise the regexp too.


ifoche:
@idelcano just a couple of comments regarding this PR:
- I think you (or me in my explanation, I don't know) inverted the font size for levels 1 and 2. The idea was to find a font size between 1st level titles and the questions, for the 2nd level title, so 1st level titles would be the biggest, after them, the 2nd level titles, and then questions and rest of things. What I see in your branch is that 1st level titles have now that new intermediate size, instead of giving it to the 2nd level titles...I think it's better if you see it in an image...

feedback_feedback

- To test the asterisk indication that a survey is not yet sent, I've used the "brute force" method :). So I disconnected the network cable from my computer, ensuring no connexion was possible (so no possibility of sending the survey). After that I create a survey, fill it in, and mark as completed. During a very little period of time I can see it in improve tab with its asterisk, but soon the asterisk disappears, even if no connexion was possible. Is it a bug? Am I missing something in the process?

Thank you

idelcano:
Done.
The first point, the problem was in diferent resolutions, i fixed it.
The second point, we talked about it in a meeting, i think ivan added it in Push_test, but i dont find it, i fixed this problem in my last commit.

PR #766, opened 2016-02-11
Closes #749, max events per orgunit/program as a setting

### LINKED ISSUE #749: Adapt code to SDK parametrized download
Adapt malaria code to the code contained in the PR #16 , that will be merged before this task begins

ivanarrizabalaga:
Closes #749

Requires https://github.com/EyeSeeTea/dhis2-android-sdk/pull/29 to be merged into *sdk: 2.22* to compile.


ifoche:
we had said 300 as the default value, and it is currently 200 as far as I can see. However, let's leave it like that (200) and wait for the client comments before changing it

PR #782, opened 2016-02-22
Productivity as a combination of attributes from orgunit and program

### LINKED ISSUE #765: Change productivity to a combination of Program Attribute, OrgUnit Attribute and Control dataElement

ivanarrizabalaga:
Closes #765


ifoche:
@idelcano @ivanarrizabalaga I'm merging #782 and #777 . That means that we need to change the server against we make the tests to hnqis-dev-ci. @josemp10 can you please move the metadata last changes (DEType and DEQuestType) to hnqis-dev-staging so we can continue using it? Thanks in advance!

PR #785, opened 2016-02-24
Dataelement conflict

### LINKED ISSUE #780: Check wether or not the ImportSummary is detecting the conflicts and put the survey in a new "CONFLICT" state in that case.

idelcano:
Closes #780
New features:
when the dataelement have a conflict:
The survey show "Conflict" text in the score.
The survey show in the feedbackfragment the value with "Conflict" text in the score
when the app push a survey with conflicts, it shows a dialog with the error, only one.

The import Summary is not return or saving the conflict, but the FailedItem table does it.
Saving(for example):
{"httpStatus":"Conflict","httpStatusCode":409,"status":"WARNING","message":"One more conflicts encountered, please check import summary.","response":{"responseType":"ImportSummaries","imported":29,"updated":0,"deleted":0,"ignored":1,"importSummaries":[{"responseType":"ImportSummary","status":"SUCCESS","importCount":{"imported":29,"updated":0,"ignored":1,"deleted":0},"conflicts":[{"object":"jIIYvQtD3zW","value":"value_not_numeric"}],"reference":"dpEkxlCkRIH","href":"https://hnqis-dev-staging.psi-mis.org/api/events/dpEkxlCkRIH"}]}}

I create the column "conflict" as boolean in the Value.class.
I save in this column if the Value has conflict, I saved it when the survey is converting to the sdk looking the sdk db in the FailedItem table.

This is needed for show the conflict in the improve tab(the survey with conflict) and in the feedback(the value with conflict)

I think other options like: Create Conflict Pojo, and save all the JSON Values, or create a table(ValueConflict), but i think add a column in the value is more fast for this feature.

Other thing. I was trying to make the test, but i can´t simulate the bad dataelement type. I am not sure if continue trying to finish the test tomorrow.

### DIFF
--- a/app/src/androidTest/java/org/eyeseetea/malariacare/test/push/PushConflictTest.java
+++ b/app/src/androidTest/java/org/eyeseetea/malariacare/test/push/PushConflictTest.java
@@ -112,8 +112,10 @@ public class PushConflictTest {
     @Test
     public void pushWithDataElementConflict(){
         //GIVEN
+        /*
         login(HNQIS_DEV_STAGING, TEST_USERNAME_WITH_PERMISSION, TEST_PASSWORD_WITH_PERMISSION);
-        waitForPull(40);
+        waitForPull(DEFAULT_WAIT_FOR_PULL);
+
         startSurvey(1, 2);
         fillSurvey(7, "Yes");
         Survey survey=Survey.findById(getSurveyId());
@@ -127,7 +129,6 @@ public class PushConflictTest {
         //Simulate the conflict dataelement
 
         //then: Survey is pushed (UID)
-        /*
         Log.d(TAG, "Session user ->" + Session.getUser());
         survey=waitForPush(20,survey.getId_survey());
 
--- a/app/src/androidTest/java/org/eyeseetea/malariacare/test/push/PushConflictTest.java
+++ b/app/src/androidTest/java/org/eyeseetea/malariacare/test/push/PushConflictTest.java
@@ -88,7 +88,7 @@ import static org.eyeseetea.malariacare.test.utils.SDKTestUtils.waitForPush;
 @RunWith(AndroidJUnit4.class)
 public class PushConflictTest {
 
-    private static final String TAG="PushOKTest";
+    private static final String TAG="PushConflictTest";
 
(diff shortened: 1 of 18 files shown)


ifoche:
thanks @idelcano 

For the test, do you think it would be very difficult to modify the answer for a given DE to manually create the conflict by changing the type of the answer? If so, maybe you can create a new push test (starting from Ivan's one) and make a new "fail push".

idelcano:
I was trying  but is not simple. I cant simulate the change of type in server. I will try more.

Anyway i fix the push fail withoutpermissions
i change the program of the survey, and the push fails becouse the user not have permision(Test ok)

(I remember you the no_permisions user never finish the pull at this moment, and the pushfailtest was failing)

ifoche:
I think I didn't explain myself very good. I said "starting from Ivan's one", not "starting from Ivan's commented one". There's only one Ivan's push test working now if my memory doesn't plays tricks on me, I was talking about that one. 

Regarding the change on the server side...you're right, you cannot change the server side, so the way to "simulate" the server side change, is to submit a wrong type (change it in the local side), and there, you should be getting the same error...don't you think?

ifoche:
@idelcano I move this task to "In progress" so, please, move it to "To test" again once the test is done, and working

idelcano:
@ifoche
I think the problem is all the dataelements have a valuetype "text" in the server.
If i send other option(like Real/simulated, or Yes/no or Number) it can be casted to "text", and the server save this value without fails...

For do this branch i tried changing a dataelement whit "text" value to "integer".
And when i send a "text"option it cant be a number and fail...

I´m not sure what can i do.


Other thing.
 If we have a one bad dataelement in one program(changing the valuetype from text to integer) this branch work for the test. (For example if one new usertest can be test the program "Pregnancy and Delivery complications" and we change one dataelement valuetype to integer).
If we pull from this program all works ok, the option set is correct(Yes/no or similar), but when we try to push, the valuetype is not correct and the error is registered.

ifoche:
ok, leave it commented for the moment, and what we probably will do is to change the program that we're using for the testing, to another one that may contain integers. That way we would be able to force the error.

PR #799, opened 2016-02-29
Fix ui feedback

### LINKED ISSUE #793: UI feedback

idelcano:
Closes #793 
One comment about the last point. I edit the Overdue row color too(I think you maybe say never assessed but want said overdue)

### DIFF
--- a/app/src/main/assets/dashboard/dashboard.html
+++ b/app/src/main/assets/dashboard/dashboard.html
@@ -154,6 +154,32 @@
     .show {
     display:block;
     }
+    #selectProgram{
+        top: -5px;
+        position: absolute;
+        width: 98.2%;
+        border-radius: 0px;
+    }
+    #tableCanvas,#pieChartContent{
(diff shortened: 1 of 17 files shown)


ifoche:
@idelcano one thing that I think I told you yesterday. I see that you chose finally to remove the tabs arrow when there's not any other arrow. There's a steady time when you press on the arrow, just before the new tab appears, that you lose the alignment. Probably the arrow addition/remove is being done later and then all the alignment is readjusted. Do you think that it is possible to avoid that effect (maybe removing/adding the arrow early in the change tab process)?

idelcano:
@ifoche  it is true, finally i could solve it using setAlpha to make the img transparent or visible.

ifoche:
another thing, can you also unify the appearance of every "ALL ASSESSMENT" filters (from every tab) to make them all look the same?

idelcano:
Done !

PR #801, opened 2016-03-01
Loads option factor via OptionAttributeValue 'OF'

### LINKED ISSUE #778: Separate in the code option from option factor

ivanarrizabalaga:
Closes #778 (requires sdk to be updated with option_attributes stuff too)

### DIFF
--- a/app/src/main/java/org/eyeseetea/malariacare/database/iomodules/dhis/importer/ConvertFromSDKVisitor.java
+++ b/app/src/main/java/org/eyeseetea/malariacare/database/iomodules/dhis/importer/ConvertFromSDKVisitor.java
@@ -245,7 +245,7 @@ public class ConvertFromSDKVisitor implements IConvertFromSDKVisitor {
         appOption.setName(sdkOption.getName());
         appOption.setCode(sdkOption.getCode());
         appOption.setAnswer(appAnswer);
-        appOption.setFactor(DataValueExtended.extractFactor(sdkOption.getCode()));
+        appOption.setFactor(sdkOptionExtended.getFactor());
         appOption.save();
     }
 
--- a/app/src/main/java/org/eyeseetea/malariacare/database/iomodules/dhis/importer/models/DataValueExtended.java
+++ b/app/src/main/java/org/eyeseetea/malariacare/database/iomodules/dhis/importer/models/DataValueExtended.java
@@ -79,7 +79,9 @@ public class DataValueExtended implements VisitableFromSDK {
             String optionCleaned=extractValue(option.getCode());
             String valueCleaned=extractValue(dataValue.getValue());
             //Yes[1]==Yes || Yes==Yes || Yes==Yes[1]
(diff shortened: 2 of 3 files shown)


ifoche:
@ivanarrizabalaga I think we're having a problem with this code (or with data in hnqis.psi-mis.org server). I can identify 2 problems at this moment:
- CompositeScores are not being translated (maybe related to DEType/DEQuestType changes)
- [These lines](https://github.com/EyeSeeTea/malariapp/blob/option_attributes/app/src/main/java/org/eyeseetea/malariacare/database/iomodules/dhis/importer/models/DataValueExtended.java#L82-84) are not properly finding the options, as now the option code does not contain the value inside. I cannot find any consequence of that failure (you'll see a lot of [this](https://github.com/EyeSeeTea/malariapp/blob/option_attributes/app/src/main/java/org/eyeseetea/malariacare/database/iomodules/dhis/importer/models/DataValueExtended.java#L87-87) warnings spread along the last part of the pull.

Can you check it and tell me if you can throw a little more light on this? First problem is probably a data problem, and second problem may not have any consequence, but I don't want to merge the code without being sure. I think you already had a user but please let me know if you need a user there to check

Thanks

ifoche:
(just in case, I updated the last comment to remove one of the 2 problems I had identified, because repeating the process with a user with less permissions it doesn't happens, so I imagine that one was related to retrieving from the server data not ready to be read by the app)

josemp10:
Regarding this:

   - CompositeScores are not being translated (maybe related to
   DEType/DEQuestType changes)

Yes, all dataelements in hnqis.psi-mis have been translated to the "new"
DETYpe/DEQuestType configuration (the optionsets and attributes have also
codes). Are the KE comp.scores working in https://hnqis-dev-ci.psi-mis.org ?

ifoche:
@josemp10 I think you missed my last comment and the edition of the previous one.

ivanarrizabalaga:
**SUMMARY**
The data from server (hnqis.psi-mis.org) implements the new model to separate Questions|CompositeScores|ControlElements but NOT the old one anymore.

**DETAILS**
Previously the separation between these 3 was based on the value for the attribute with code 'DEQuesType'.

In 'https://hnqis.psi-mis.org' this attribute is the one with UID = 'RkNBKHl7FcO'

In order to check if the dataelementAttributeValue corresponds to a CompositeScore or a Question you had to check first the expected value for the COMPOSITE_SCORE option which was 93.

If you look for a combination of attribute:'RkNBKHl7FcO'(DEQuesType) + value:'93' there are NO records so with the old conversion there are no CompositeScores.

The new conversion looks ok, looking for DEType + 93 in dataelementattributevalue returns 74 records.

ivanarrizabalaga:
@ifoche , after pulling the branch everything is fine:
 - CompositeScores are loaded
 - I've added a new commit to ensure that value options are filled only with an exact match between the option.code and the datavalue.value

ifoche:
ok @ivanarrizabalaga , thank you very much!

PR #808, opened 2016-03-03
The ROOT_NODE_CODE is 1 not 0 at this moment in the server... It fix the

### LINKED ISSUE #804: Score table in monitoring not properly working

idelcano:
closes #804 

The problem was: The root CS score changed from 0 to 1 (maybe long time ago).
the mainscore works now.
The monitortab works now with the correct pulled data.

### DIFF
--- a/app/src/androidTest/java/org/eyeseetea/malariacare/test/pull/PullTranslationSDKToAppDBTest.java
+++ b/app/src/androidTest/java/org/eyeseetea/malariacare/test/pull/PullTranslationSDKToAppDBTest.java
@@ -7,7 +7,13 @@ import android.support.test.rule.ActivityTestRule;
 import android.support.test.runner.AndroidJUnit4;
 import android.util.Log;
 
+import com.raizlabs.android.dbflow.sql.builder.Condition;
+import com.raizlabs.android.dbflow.sql.language.Select;
+
 import org.eyeseetea.malariacare.LoginActivity;
+import org.eyeseetea.malariacare.database.iomodules.dhis.importer.CompositeScoreBuilder;
+import org.eyeseetea.malariacare.database.model.CompositeScore;
+import org.eyeseetea.malariacare.database.model.CompositeScore$Table;
 import org.eyeseetea.malariacare.database.model.OrgUnit;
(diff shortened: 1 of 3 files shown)


ifoche:
@josemp10 was this metadata change something that you wanted or is it just a missing CS (I think so)? @idelcano, I will not merge this code until @josemp10 answer, as it may be possible that it is simply a CS that is not in the server and it should be. BTW. Just to tell @josemp10. You're testing against hnqis.psi-mis.org, isn't it?

idelcano:
I was testing against hnqis.psi-mis and against hnqis-dev-ci
In hnqis.psi-mis the CS in the pulled DB are all like:
hierarchical_code  label uid order_pos
"1"	"HIV"	"PMNfHllbkDy"	"0"
"1.1"	"HIV Testing and Counseling"	"gZLPkawuht2"	"1"
"1.2"	"HIV Care and Treatment"	"Ozg2PzNmKJO"	"2"
"1.3"	"Work Environment"	"YHNyxjvpG3m"	"3"
But, i look it:
hierarchical_code  label uid order_pos
"0"	"ICM"	"XgjACuQSvun"	"0" (it is child of the next)
"1"	"Inquiry and Physical Examination"	"mmfP19C6Qxm"	"1" (it is the parent)
"1.1"	"Introduction"	"YolOXfWtsRB"	"2"
"1.2"	"Inquiry"	"mdZjidBIBa6"	"3"
"1.3"	"Physical Examination"	"GICXKJjx1Sv"	"4"
"1.4"	"RDT Procedure"	"F40zWrhpJL2"	"5"

And if i checks the the IMCI feedback the app crash when calculate the MainScore looking the CompositeScores

ifoche:
@idelcano, that explains everything. I was always testing with IMCI, that is properly set. @josemp10 is changing the metadata in hnqis.psi-mis.org (@josemp10 would it be possible to do that too in hnqis-dev-ci server?) @josemp10 thinks that the last not-very-lucky tries in hnqis failed because the hierarchichal code was not set. He will tell us more about this issue when he updates the metadata

ifoche:
BTW, I will probably close this issue without merging it, if @josemp10 can test that effectively, a metadata change would solve everything.

PR #809, opened 2016-03-03
Fix plan hide problems

### LINKED ISSUE #805: tab switching time very long when planning tab is hidden

idelcano:
Closes #805


idelcano:
(this PR solved #813 issue too)

ifoche:
@idelcano I think that a future refactor of this code would be needed. I'm going to approve it now because we need it, but we'll come back on this later

Issue #814, opened 2016-03-03
Yellow and green colors of the improve list are not visible (they don't contrast with the background)


ifoche:

ifoche:
It would be fin to use a darker yellow and green, and maybe use a bold font if it's not yet bold

PR #818, opened 2016-03-04
Monitor colors

### LINKED ISSUE #812: Change monitoring pie chart colors to match the feedback ones

idelcano:
Closes #812 
Comment:
I only change the pie chart colors(and pie table colors), not the all assessment table colors(target blue and  Assessment undertaken green), becouse i think it is other view.

Ok other problem: i set the colors from feedback, but i will change this colors in #814 .
Not take time, but maybe is good not forget it.

### DIFF
--- a/app/src/main/assets/dashboard/dashboard.html
+++ b/app/src/main/assets/dashboard/dashboard.html
@@ -117,8 +117,9 @@
     margin: auto;
     border-radius: 50%;
     height: 100%;
-    background-color: #81980d;
+    background-color: #adda19;
     color: white;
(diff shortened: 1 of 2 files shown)


ifoche:
@idelcano I think we talked about introducing the colors in the html/js part by getting them directly from the already defined color in the color.xml file and not by editing them by hand. I'm going to merge it as this is something that we need, but I won't delete the branch, so would you please add that part in this branch and do another PR? (I will leave the task open) Thank you

idelcano:
Yes, we talk about it but after the PR. 
I will do it.

PR #827, opened 2016-03-05
Generate last commit commit in pre push

### LINKED ISSUE #798: Introduce build number (or commit number) in the about text
See if we can do it directly with git variables. Otherwise, maybe buddybuild can introduce it by himself. The last and worst option would be to add it as a gradle variable manually modified

idelcano:
Closes #798
@adrianq  @ifoche   
Finally i did this solution:

I created a bash script, the script is running in all git pre-push.
This scripts checks if you are pushing against a selected repository (looking in the array "repositories" of the script). 

The script checks:
-repository
-last commit is not an automaticall last commit (in our local repository).

And does:
git reset
git add $filePath
git commit -m "$commitString"	

The funcionallity is...
when you make a push, 
you get the lastcommit hash and create the last commit if is necessary.

To "install" the script is simple:
We need add the pre-push file in malariapp\.git\hooks\ (without extensions, unzip the file and copy past).
[pre-push.zip](https://github.com/EyeSeeTea/malariapp/files/160029/pre-push.zip)

Before commit the last commit, i added git reset
for be sure don´t commit more files.

Note: In my historial of commits in this branch, i have some "Added the last commit number" becouse i tried "git reset" works, and i added a little change, and i did 5 pushes. I changed the string for "Added the last commit hash" in the script.  

I was looking more solutions, for example using git attributes and git filter, but i can´t do it work for our purposes:
For example i can change the local file with the lastcommit hash on git checkout or in git add but i can´t commit and require some local configurations, and git ignore this changes, and it dont work for buddybuild only in local enviroment).


ifoche:
@idelcano let's talk about it later whenever I come back to Europe. I see some problems to the sustainability of this, and I think that this would insert a lot of commits in the history (it will duplicate the history by putting a commit after each commit). I leave it here without merging it and we'll see it later

idelcano:
Yes, i was not sure about this solution was good.  Commit the last commit only in every push.

But, i thinking about it.
Maybe in the buddybuild pre build script we can add a script to overwritte the commit.txt with the real last commit of the active branch.. 

Adding some like(bash scriptting):
#!/usr/bash
filePath="app\src\main\res\raw\commit.txt"
commit=`git log  -1 --pretty=format:"%h"`
echo $commit>$filePath

Looking your buddybuild scripts i think it should work....

And for remember in the future:
I was learning about git filters clean and smudge, but it only works when you commit or do checkout(with changes) of a file. And i lost time trying do something with that.
I tried other things  as using $ID$(replaced by git for a object hash of a file), but it is not valid for ours purposes, some people was adding the date but i think it dont fixed nothing.

idelcano:
I Added in the  buddybuild post-clone script two commands.(read the last commit, and overwritte commit.txt files on the proyect files). And it works.

idelcano:
@ifoche 
The post-commit and post-checkout works in the development side.

But i can't run the post-checkout hook in buddybuild.(the script is created and works, but the hook never works).

The last commit is updated now, because i added the script execution in buddybuild_postclone directly.

Issue #845, opened 2016-03-09
When parent question hide its children, ask for confirmation to delete the children values


ifoche:

idelcano:
@ifoche sorry, i am having problems with this task.
The problem was:

I need create a dialog from a Activity(for example from DashboardActivity, it is the activity open when survey fragment is display), becouse the radiolistener not have activity  context to call the dialog.

But i have two problems.
The first is the dialog, is showed too late.. 
It is like in other thread. 
When the dialog is showed in the screen, all the actions was ran time before.
And the second, i need get the response of the dialog from a static variables and i need stop the onCheckedChanged action.


I was trying setting a clicklistener in the radiobuttons button , but the click is running before onCheckedChanged.

I did for try to understood the problem:
I try to do:
show dialog
thread sleep(isdialoganswered?)
actions.
But the dialog never was showed.

if the secuence is:
show dialog
actions

The app do:
actions
...
show dialog

I was trying now using other listeners for the buttons if the radiobutton question has childs.
For example, i´m trying, setOnTouchListener (here i should the dialog) and setOnClickListener(if the dialog was showed, run it and click the button)
The order in execution are Touchlistener CheckedChangedListener and ClickListener.
I need show the dialog in Touchlistener, but wait for it before run the others listeneres.

This is taking me so many time trying ways to show the dialog and get the values.

Do you have any idea how can i do it?
Do you prefer i change of issue and still think about it?
Or is it more important than the others and it needs be finished it first?

thanks!

ifoche:
@idelcano I need you to upload whatever you could have about this issue to a branch and move to another issue. And please first move back again to the parent-questions progress wheel task as it is not yet working (I'll explain there why not). Just to give you an idea about the pressure that we have, tonight we're releasing the last version that should contain every bug solved, so please if you find problems today working on an issue, comment it and move quickly to the next one, I'll be around to try to solve those you leave behind.

idelcano:
@ifoche the branch is (add_confirm_before_close_childquestions)
https://github.com/EyeSeeTea/malariapp/tree/add_confirm_before_close_childquestions

PR #849, opened 2016-03-09
Progresbar wheel on match

### LINKED ISSUE #844: Add a progress bar (wheel) when a parent question is selected

idelcano:
Closes  #844


ifoche:
@idelcano this is not working due to a misunderstanding problem. Maybe I didn't explained it properly. When I talk about parent questions, I'm not talking about match ones (btw, I'm not getting any progress wheel on match either) I was referring the questions that show/hide other questions. For example, in family planning, a parent question could be "Does this facility offer contraceptive counseling?". If you press yes there, the children questions appear. Please confirm me that you've now understood what I'm talking about. @adrianq can you explain it in Spanish to @idelcano by hangouts? Thank you very much

idelcano:
Yes i understood and it works for my, i will revise all.

ifoche:
ok, so if it works for you just let me check it again just in case

idelcano:
no wait!, it works, but bad.
The problem is:
i tried in emulator and it is so fast. the progress wheel is showed AFTER the child questions was showed.
And i need do it before.
I will review it.

ifoche:
ok, thank you for taking care of this

idelcano:
The problem is the same as in  #845

My changes in the UI are showing after the hide/unhide childs.

I don´t understand why i can´t look my ui changes  before the UI list was reloaded.

ifoche:
I'm going to implement the workaround of introducing a different image for selected+pressed and for non-selected+pressed. That way, although we won't see the wheel, we'll get feedback of the click. However, please continue tomorrow working on this to guess how to solve it

idelcano:
Ok @ifoche  thanks. i will continue tomorrow. I lost all the day with this and the other issue, i look your other PR, great solution (but the dialog is showed always, when you push yes or not when the childs are visibles or not) I think the dialog should be showed only if the childs are visibles.

idelcano:
@ifoche  sorry i was wrong yesterdey, the dialog is showed always in your solution of the other PR, it works ok!. 
I think you are working on this issue, but  i take a look.. 

I uploaded my branch, the progressbar is showed now. A little freeze, but showed when the user press a parent radiobutton and hide when the list was loaded.

ifoche:
@idelcano it's still not working from a functional perspective. I've tested it and it takes sooooo long in a real device, and the progress bar don't even move at all. We need to analyze it better. Please, I've merged it with development. Whenever you have time it would be perfect to come back to this. It's not going to be tested in the field in Tanzania, but I would like to include it for the final R14 release (probably around the end of next week)

ifoche:
@idelcano I'm traveling tomorrow all the day, but maybe we can meet on friday and review this task? Can you update me with the state of this task?

idelcano:
@ifoche 
The problem now is the list method .notifyDataSetChanged().

I try to do the DB operations and score operations in background(in asyntask), and the progressbar looks good, but when we call .notifyDataSetChanged() the progressbar is gone, and the list needs 1 o 2 seconds more to refresh the items.
I will look about the notifydatasetchanged method.
I think i can finish this task today maybe.

About the friday meeting, do you know when?
If you want we can meet the saturday.

ifoche:
@idelcano I was thinking on meeting at 16h, I just add you to the calendar event.

idelcano:
@ifoche i uploaded this branch. I think it is more good now. It looks good in the emulator. I want try it in real device but, i dont why, all the child questions was unhidden in my phone.

ifoche:
@idelcano I'm not approving this yet. I want to talk to you and show you the behaviour in a real device, so we can talk about it and see if there would be a solution to improve the user experience a little bit more.

idelcano:
I didn't find a better solution
But i tested in a real device, after change the language in my phone.
i am really unhappy with the performance. I think this is not valid. In the emulator looks really better.. But in my phone... All the steps are too slow.

Maybe we need other solution like use a static image, or at least not a semi-transparent wheel. Or a progressdialog.

ifoche:
Closing this PR as it finally didn't reach the required performance in real devices

PR #876, opened 2016-03-11
Added try{}catch to prevent the app crash if the user don´t wait for the

### LINKED ISSUE #875: crash if changetab when the surveyfragment is loading
I was testing the app and i looked this:

The surveyFragment is loaded in asyntask . 
If you change the tab when is loading, the asyntask try to use null data, and the app crash.

A simple try{}catch solve it. I will do the PR.

idelcano:
Closes  #875 
@ifoche 
I was trying the app and i look this "bug".
Maybe you want do PR, maybe not.
I hope it´s not too late.


ifoche:
@idelcano please just remember that catching Exceptions with "Exception e" maybe is not the best practice. I'm going to accept this change for this pilot, but one of the first code managements in the next release I think that it should be to make more concrete the type of exception that is being captured in the try-catch-finally statement. Thanks for the work!

PR #887, opened 2016-03-21
Fix all assessement typo

### LINKED ISSUE #882: Planning tab "ALL ASSESSMENT" should have a final S

idelcano:
Closes #882


ifoche:
@idelcano why are you using a string in uppercase instead of setting it to uppercase by calling toUpperCase() function? Is there any special reason for changing that?

idelcano:
No, the reason is the string is always uppercase.
That string is only used in the filters. If we change it we need remember
set as uppercase all the places when we used that string.

I changed "all assessment" string used in the planning tab by the all
assessments string used in the other filters.

2016-03-21 12:42 GMT+01:00 Ignacio Foche <notifications@github.com>:

idelcano:
@ifoche it is done.

idelcano:
@ifoche  it is done

PR #920, opened 2016-04-07
Pull controldataelements

### LINKED ISSUE #913: Pull the controldataelements uid

idelcano:
Closes #913
It needs [sdk branch](https://github.com/EyeSeeTea/dhis2-android-sdk/tree/pull_all_programDataelements)

### DIFF
--- a/app/src/eds/res/values/strings.xml
+++ b/app/src/eds/res/values/strings.xml
@@ -26,17 +26,4 @@
     <!-- default server url -->
     <string name="login_info_dhis_default_server_url" translatable="false">https://eds.psi-mis.org</string>
 
-    <!-- Control Data Elements (This data should be changed once these uid are uploaded to the server)-->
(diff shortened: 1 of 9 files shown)


ifoche:
@idelcano why are you creating a table ControlDataElement inside our DB? If I'm not missing something, this is a bad idea:
- We already have a DataElementExtended pojo that points to DataElements. If you want to classify the different type of DE with different objects, I think it's a good idea but you can simply extend from the DataElementExtended pojo, can't you?
- As you're introducing the table decorator with our DB name, you're storing that as a new table in our DB. Do we really need it? The control dataElements are translated in our DB with different meanings in our DB. 

I really can't see the need of that pojo. Let's talk about it today

idelcano:
@ifoche We forget talk about it.
I think this kind of data needs be stored in db. 
And it is different than the others pojos.
It is like a question but not is it...
Can we change the name for something like surveyControlData, surveyMetadata ...?

PR #921, opened 2016-04-07
Added default value "https://" in login

idelcano:


ifoche:
@idelcano I think there's a misunderstanding here. They want to change the default server for EDS. If I'm not wrong, you're changing hnqis here.

PR #945, opened 2016-04-13
Pulls just the last event from server for each orgunit,program

### LINKED ISSUE #929: Pull last event for program/orgunit
Since EDS will easily have 1000 events for some users it is required to make it work under such conditions.

We need to fetch the last survey (event) from the server for each program/orgunit.

ivanarrizabalaga:
Closes #929


idelcano:
@ivanarrizabalaga  @ifoche  

The changes in ConvertFromSDKVisitor file in the commit 
[6ebb4f8251902a44330969bd6fc5b28643e6801c](https://github.com/EyeSeeTea/malariapp/pull/945/commits/6ebb4f8251902a44330969bd6fc5b28643e6801c) should be removed before merge.

(Except "survey.setScheduleDate(sdkEventExtended.getDueDate());")

Those dates are pulled from the control dataelements in next steps.

We set the eventdate in those dates to prevent a null value. 

And we use the eventDate in all of them, because eventDate is not changed by the server.

PR #967, opened 2016-04-21
Custom tabs compositescore tests

### LINKED ISSUE #960: ScoreRegister unit tests

idelcano:
Closes #960


ifoche:
I return this PR to ready in order to work on it later on. Please not this is a very low priority task and would consist on update these tests with latest code. It may require to completely rewrite them if code is so different (but I don't think so, as score code has remain almost untouched since very long time).

PR #997, opened 2016-05-11
Push username instead of user UID

### LINKED ISSUE #996: Push username

ivanarrizabalaga:
Closes #996


ifoche:
Closing this PR. The username is unique but it cannot be retrieved using the DHIS2 API, so the only unique field that last is UID

PR #1028, opened 2016-05-23
Added: different text sizes by screen and language

### LINKED ISSUE #1021: if screen resolution is very low, extra large font size is not offered

idelcano:
closes #1021


idelcano:
I finished the task as we said in the meeting.

ifoche:
@idelcano this code, at least in my case, is always removing the extra large option, no matter what screen size you have. Did you test it on a large screen?

ifoche:
@idelcano there's a bug on isShowedLargeText() method. The returned value is forced to be false just before returning it...so the method itself is useless. I'm fixing it.

PR #1036, opened 2016-05-26
Bug fix composite score bug

### LINKED ISSUE #1034: Application crash adding registers in Scoreregister from survey session
The Application crash when tries to add registers in the scoreRegisters from the Survey saved in session.

This problem appears when  the user open one survey feedback and some surveys are pushing in backgraund at the same time.

idelcano:
Closes #1034


idelcano:
@ifoche  i finished this branch. 

But the PR( #1037) against developemnt is not updated.

PR #1056, opened 2016-05-30
Feature add list filter option

### LINKED ISSUE #1049: In the improve tab, show all the assessments in the DB, not only the last one for each OU

idelcano:
Closes #1049


ifoche:
@idelcano I'm trying to guess why this code is not doing what it was supposed to do. Do you have any not-uploaded commit in this branch?

ifoche:
sorry @idelcano, too much merges with the same DB. I've reinstalled the app and now it seems to work

PR #1159, opened 2016-08-09
Feature planning view

idelcano:


app/src/main/java/org/eyeseetea/malariacare/database/utils/planning/PlannedServiceBundle.java
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Created by idelcano on 08/08/2016.
+ */
+public class PlannedServiceBundle {
+    List<PlannedItem> plannedItems = new ArrayList<>();
+    List<OrgUnit> orgUnits;
+    List<Program> programs;
+
+    public PlannedServiceBundle(){
```
adrianq: @idelcano I don't think this constructor is needed (unless you are thinking about adding something else later on). Just initialise the variables. Also the variables should be private if we have getter/setter.

app/src/main/java/org/eyeseetea/malariacare/fragments/PlannedFragment.java
```diff
@@ -92,8 +102,8 @@ public void onActivityCreated(Bundle savedInstanceState) {
 
     }
 
-    private void prepareUI() {
-        this.adapter = new PlannedAdapter(plannedItems,getActivity());
+    private void prepareUI(List<PlannedItem> plannedItems) {
+        this.adapter = new PlannedAdapter(this.plannedItems,getActivity());
```
adrianq: @idelcano this is a little bit confused. Is plannedItems a parameter or variable of the class?

app/src/main/java/org/eyeseetea/malariacare/DashboardActivity.java
```diff
@@ -819,4 +815,43 @@ public void run() {
...
+        changeLayout = (LinearLayout) findViewById(R.id.dashboard_planning_init);
+        changeLayout.setVisibility(View.VISIBLE);
+        plannedFragment = new PlannedFragment();
+        plannedFragment.setArguments(getIntent().getExtras());
+        replaceListFragment(R.id.dashboard_planning_init, plannedFragment);
+    }
+
+    private void initOrgUnitFragment(OrgUnit orgUnit) {
+        //hide plannedFragment layout and show plannedOrgUnitsFragment
+        LinearLayout changeLayout = (LinearLayout) findViewById(R.id.dashboard_planning_init);
```
adrianq: @idelcano What is going on here? Take a look at these lines. Too crazy to merge it :)

adrianq:
@idelcano I am OK with merging this PR even if it is not polished and fully accomplished (specially considering that @ifoche is going to start the custom tabs huge merge) however there are a few things that would be good to improve/fix before merging. I have left a few comments, can you take care of them and in general double check the code?

PR #1162, opened 2016-08-15
Bug score register children questions

idelcano:
Modifies the children num/denum workflow.
(fix the required functionality)

### DIFF
--- a/app/src/main/java/org/eyeseetea/malariacare/layout/score/ScoreRegister.java
+++ b/app/src/main/java/org/eyeseetea/malariacare/layout/score/ScoreRegister.java
@@ -207,17 +207,10 @@ public class ScoreRegister {
         }
         Value value = question.getValueBySurvey(idSurvey);
 
-        //when a child question is compulsory but is not active, its ignored
-        if(question.getCompulsory() && question.hasParent()){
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/eyeseetea/malariacare/layout/score/ScoreRegister.java
```diff
@@ -206,14 +206,17 @@ public static Float calcNum(Question question, float idSurvey) {
             return null;
         }
         Value value = question.getValueBySurvey(idSurvey);
-        //Returns null if the question will be ignored(not compulsory, and not answered or child with inactive parent questions)
+
+        //when a child question is compulsory but is not active, its ignored
+        if(question.getCompulsory() && question.hasParent()){
```
adrianq: @idelcano what is this? Three 'if' one inside the other and finally a 'return null'. Can this code be improved? Same for the 'if' below

idelcano:
@adrianq  I re-thought this, and finish in more cleaner way. And i solved a little bug too(on repaint a row the denum didn't shown). 

Note: This branch needs the children compulsory branch, i did push with this commit by mistake, i reverted that using "git revert".

PR #1196, opened 2016-09-21
Fix double dialog when click on parent question options

### LINKED ISSUE #1192: unchecking a parent bug.
When unchecking a parent with answered children, the app shows a dialog two times. The first one hide all the children but the second shown the children again.

idelcano:
closes #1192


ifoche:
Looks like there are still problems with this. Double dialog may be solved, but when you click on ok, it's not deleting all the children's values. I've sent you an email with a video showing it

idelcano:
I fixed it.

idelcano:
@ifoche  i finished it.

PR #1357, opened 2017-04-10
Feature survey fragment progressbar and messages

### LINKED ISSUE #1344: Introduce Progress bar on back button pressed (inside a survey)
When back is pressed inside a survey, if the survey contains 1000-2000 questions, it last for a very long time. So we need to add a progress bar to let user know things are happening (all the calculations for determining if the survey has all mandatory questions filled or not)

idelcano:
Closes #1344

Only for test this feature if the backbutton dialog is shown too fast:

I used the program HIV with all the answers all parent/child questions checked.
And i modified a local DB setting all the questions as compulsory to test it better.


app/src/main/java/org/eyeseetea/malariacare/data/database/model/Survey.java
```diff
@@ -551,30 +549,39 @@ private long countNumOptionalQuestionsToAnswer() {
...
-            }
-
-        } else if (surveyAnsweredRatio.getCompulsoryAnswered() == 0) {
-            this.setStatus(Constants.SURVEY_IN_PROGRESS);
-        }
-
-
-        //Saves new status & completion_date
-        this.save();
+        GetSurveyAnsweredRatioUseCase getSurveyAnsweredRatioUseCase = new GetSurveyAnsweredRatioUseCase();
```
xurxodev: @idelcano an UseCase can not be invoked from data layer, Use case should be invoked only from presentation layer. On this way updateSurveyStatus has more of one resposability. A possible option is from where we invoke updateSurveyStatus first we calculate SurveyAnsweredRatio and depending on the result invoke updateSurveyStatus with the desired state

idelcano:
@xurxodev  thanks for your comments. It's fixed

PR #1386, opened 2017-04-14
Fix locale date formats

### LINKED ISSUE #1385: Date question type has invalid format

idelcano:
Closes #1385
The date tipe wasn't invalid, but we was saving and pushing the default locale format and this is incompatible with the server in some cases.

The value is now pushed and stored with the server format, but is displayed with the locale format in the views.

### DIFF
--- a/app/src/main/java/org/eyeseetea/malariacare/domain/entity/DateQuestion.java
+++ b/app/src/main/java/org/eyeseetea/malariacare/domain/entity/DateQuestion.java
@@ -1,0 +1,120 @@
+/*
+ * Copyright (c) 2017.
+ *
+ * This file is part of QA App.
+ *
+ *  Health Network QIS App is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation, either version 3 of the License, or
+ *  (at your option) any later version.
+ *
+ *  Health Network QIS App is distributed in the hope that it will be useful,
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/eyeseetea/malariacare/domain/entity/DateQuestion.java
```diff
@@ -0,0 +1,120 @@
...
+
+import org.eyeseetea.malariacare.database.iomodules.dhis.importer.models.EventExtended;
+import org.eyeseetea.malariacare.database.utils.PreferencesState;
+import org.eyeseetea.malariacare.domain.exception.DateFormatException;
+
+import java.text.ParseException;
+import java.util.Date;
+import java.util.Locale;
+
+public class DateQuestion {
```
xurxodev: @idelcano this is not a entity, It looks more like a utils class with all methods static that we have to try to avoid. A entities are domain concepts for example value, survey, question but not exists yet. In this case the responsability to convert date formats is of adapters, remember that adapters translate information from and to domain. We still do not have very defined domain and adapters according to clean architecture but conversions with current state can be done from ConvertToSDKVisitor, ConvertFromSDKVisitor and from activity, fragment or adapter that render the date on the screen.

idelcano: @xurxodev understood, fixed.
Thanks for the explanation.
I tried to do something similar to the Positive or Zero class in pictureapp, but it is finally different

PR #1483, opened 2017-06-06
Fix uploaded by user control dataelement value

### LINKED ISSUE #1480: Push user Control DE sending UID instead of username

idelcano:
Closes #1480


ifoche (approved):
thanks @idelcano

PR #1493, opened 2017-06-12
Remove long click listener from list of DashboardSentFragment

### LINKED ISSUE #1492: Remove long press action on feedback
On feedback list, if you make a long press on an element it takes you to assess module and open that survey in read-only mode. That behavior must be removed.

manuelplazaspalacio:
Close #1492


ifoche (approved):
thanks @manuelplazaspalacio

PR #1494, opened 2017-06-13
Translating untranslated strings.

### LINKED ISSUE #1488: Untranslated strings
There are 2 strings that are not translated because they exist in POEditor but we're not using the translated strings but hardcoded ones. These are:
```
<string name="dialog_info_ask_for_completion">Are you sure you want to complete %s? You cannot undo this action</string>
```

and

```
"Are you sure you want to exit the app?"
```

Apart from those 2 other 2 strings ("Reschedule" and "Due date") are directly hardcoded in xml files and don't have an alternative in POEditor (assessment_planning_header.xml)

We need all of them to be translated

manuelplazaspalacio:
Close #1488


ifoche (approved):
thanks @manuelplazaspalacio

PR #1495, opened 2017-06-15
Bug never assessed lorem

### LINKED ISSUE #1490: Lorem ipsum on never assessed surveys reschedule date
If you go to reschedule screen on an OU for which there are "never assessed" surveys, they get "lorem ipsum" instead of the scheduled date (which should be simply void) and doesn't allow to reschedule

manuelplazaspalacio:
Close #1490


ifoche (approved):
thanks @manuelplazaspalacio

PR #1502, opened 2017-07-18
Modify sdk provider

### LINKED ISSUE #1501: HNQIS 1.2.4 provider conflict with GPlay 1.2.3
HNQIS 1.2.4 contains a provider conflict with Google Play version (1.2.3). Both should be able to coexist given that the package id for GPlay version is  org.eyeseetea.malariacare.hnqis_ng and the package id for buddybuild (v1.2_hnqis) version is org.eyeseetea.malariacare.hnqis_ng_bb. Looks like the problem that we already solved with the SDK.

Installation failed with message Failed to finalize session : INSTALL_FAILED_CONFLICTING_PROVIDER: Package couldn't be installed in /data/app/org.eyeseetea.malariacare.hnqis_ng_bb-1: Can't install because provider name HNQIS_DHIS2_DEFAULT_AUTHORITY (in package org.eyeseetea.malariacare.hnqis_ng_bb) is already used by org.eyeseetea.malariacare.hnqis_ng.
It is possible that this issue is resolved by uninstalling an existing version of the apk if it is present, and then re-installing.

idelcano:
Closes #1501


xurxodev (approved):
thanks @idelcano

PR #1527, opened 2017-08-04
order with alphabetical order the  filter querys

### LINKED ISSUE #1517: Re-order all dropdowns by alphabetical order (except OptionSets which have their own order set on server)

idelcano:
Closes #1517


xurxodev (changes requested):
@idelcano when from a survey open tab list as a special drop-down should be ordered alphabetically, can you review it, please?

idelcano:
thanks @xurxodev , I reviewed it, but the tabs already have a order field is not necessary order it by alphabetical order.
Anyway the query was wrong, i solved it adding a "orderby order field" in the query to get the tabs.

xurxodev (approved):
thanks @idelcano

Issue #1519, opened 2017-07-28
Refine all app design following mockup


ifoche:
As a guide, [this mockup](http://web2.beckolsystem.com/hnqis/#/screens/d9978ee8-0c5a-4d1e-9629-9d294141c92c) aims to provide an idea of the intention in terms of desired design changes

Another alternative design is [this one](http://web2.beckolsystem.com/hnqis-2/#/screens/d12245cc-1680-458d-89dd-4f0d7fb22724)

idelcano:
@ifoche i created this guide to simplified the task, maybe you can delete some points if you think is not important at this moment

General:
- [x] filters style 
- [x] Remove action bar back button

Fragment Planning:
- [x] remove header
- [x] plan fragment colours and fonts and listview row
- [x] plan per orgunit colours and fonts and listview row
- [x] Change button

Fragment assess:
- [x] general cosmetic changes
- [x] listview row
- [x] Create pie instead of survey % complete
- [x] Change button
- [x] Add filters

Fragment improve:
- [x] general colours and fonts
- [x] listview row

- [x] Reschedule dialog

PR #1528, opened 2017-08-07
Cosmetic refine all app design

idelcano:
Closes https://github.com/EyeSeeTea/malariapp/issues/1519
(it is not finish)


xurxodev (changes requested):
@idelcano I have found some errors:
- [x] Background of tabs should not be dark
- [x]  According to design text for all programs filter should be All Assessed
- [x]  Float action button for creating new survey should have normal size (no mini)
edited by @idelcano : I think is better ignore this design and make it like the other floating buttons
- [x]  For lists now exists a modal menu with actions but when I click on an item a default action should be invoked, this work but with this design issue there are errors
Plan-> add or edit (not work)
Unsent -> edit (work wrong)
Sent -> feedback (this work ok)
- [x]  Font in Headers for sent and unsent list and action plan should have a size bigger and more padding its container
- [x]  Action bar in action plan fragment should contain HNQIS - $module
$user
edited by @idelcano : This change is done in the next PR
- [x]  In action plan fragment the color for fabs (share) should distinct than headers of the content

xurxodev (changes requested):
@idelcano There are some points wrong yet:
- Click on unsent and sent list not work well, first row not execute default action and the next rows open previous row

xurxodev (approved):
thanks @idelcano I have fixed the last bugs as we have talked about, so I can advance reviewing the next pull request

PR #1554, opened 2017-09-18
Feature observation and action plan layout

### LINKED ISSUE #1553: Implement Observation & Action Plan layout
Following [the mockup](http://web2.beckolsystem.com/hnqis-2/#/screens/555dfab6-6e41-4fb0-98cc-4469818518fb), implement the screen layout, entered by using a button in feedback saying "Action Plan". This issue is not about the content but the layout. So the sub-tasks would be:
- [x] Add the button to access the obs & action screen from the feedback screen
- [x] Implement the layout like in the mockup, leaving the edit texts empty (nothing about DB here)
- [x] Make the edit text height variable, starting from 5 lines and expanding while the user writes (so create this new component)
- [x] For the action part, include the 2 dropdowns with the following options & show/hide logic:
1. Refresher training during SSV
   This deploys a secondary dropdown
1a. Watch video
1b. Practice proper insertion technique at time of visit
2. Group refresher training at a later date
3. Plan CME activity
4. Demonstrate technique with client
5. Other
   This shows a textfield to specify the action plan
- [x] Add the share floating button with 2 suboptions: HTML & TXT (do not implement the sharing action)

idelcano:
Closes #1553


xurxodev (changes requested):
@idelcano there are design errors according to mockup:
- [x] add padding in the header with text Observations & Action plan according to mockup
- [x] Text to left to percentage is not equal to mockup
- [x]  Gaps and action Plan EditTexts has cursor in the middle and should have at the start  
- [x]  Review bold texts (black and white)
- [x]  Today text has no sense, change to competed (review by @ifoche )
- [x]  When option with text other is selected in the last dropdown, the EditText that is visible the height is too big
- [x]  Line dividers is not equal to mockup
Can you review it, please?

xurxodev (approved):
thanks @idelcano

PR #1558, opened 2017-09-18
Feature add checkbox on feedback fragment

### LINKED ISSUE #1555: Include a checkbox to show all the feedbacks vs only last one
The checkbox would be placed on top of the header, under the filters, and would read something like "Show only last feedback per Program/OU". Activated by default, it would have the current behavior for activated, or deploy all the previous cases in case it's not

idelcano:
Closes #1555


xurxodev (changes requested):
@idelcano I find some errors:
- The behavior is the other way around. When checkbox is check only should be visible last feedback 
- checkbox should be check by default
- when checkbox is check sometimes navigate to survey and back with icon the list is empty

xurxodev (approved):
thanks @idelcano

PR #1559, opened 2017-09-18
Added critical steps missed query

### LINKED ISSUE #1556: Retrieve the list of "critical steps missed"
The feedback that will be sent by email will contain the list of "critical steps missed". This means the list of mandatory questions whose score is 0 (mandatory questions failed). This task is for including the DB query that provides that list (without front-end to show it)

idelcano:
Closes #1556


xurxodev (approved):
thanks @idelcano

PR #1566, opened 2017-09-19
Feature retrieve ob and action plan values

### LINKED ISSUE #1562: Retrieve Obs&Action Plan values from DB
When the Obs&Action Plan of a given feedback is opened, we need to check in the DB if there are already values for that survey in the DB. If so:
- If the status of the Obs&Action Plan is other than IN PROGRESS, it would open in read-only mode
- Otherwise, it would open in read-write mode

In any case, the textfields and dropdowns will be filled with the values from the DB

In case no previous data existed, a new Obs&Action Plan object is created in the DB linked to that survey

idelcano:
Closes #1562

### DIFF
--- a/app/src/main/java/org/eyeseetea/malariacare/fragments/PlanActionFragment.java
+++ b/app/src/main/java/org/eyeseetea/malariacare/fragments/PlanActionFragment.java
@@ -105,7 +105,7 @@ public class PlanActionFragment extends Fragment implements IModuleFragment {
         initSpinner(llLayout);
         initFAB(llLayout);
         initBackButton(llLayout);
-        if(mObsActionPlan.getStatus().equals(Constants.SURVEY_IN_PROGRESS)){
+        if(!mObsActionPlan.getStatus().equals(Constants.SURVEY_IN_PROGRESS)){
             setReadOnlyMode();
         }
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/eyeseetea/malariacare/fragments/PlanActionFragment.java
```diff
@@ -330,58 +351,47 @@ public boolean onBackPressed() {
...
                 if (selectedItem.equals(options[0])) {
                     mObsActionPlan.setAction1(null);
-                    mCustomActionOtherEditText.setText("");
-                    secondaryActionSpinner.setSelection(0);
-                }else{
+                } else {
                     mObsActionPlan.setAction1(selectedItem);
                 }
-                if(mObsActionPlan.getAction2()==null || !mObsActionPlan.getAction2().equals(selectedItem)) {
+                if (lastItem == null || lastItem.equals(selectedItem)) {
```
xurxodev: after merge I do not know if it makes sense this conditions with an empty body, can you review it please?

xurxodev (changes requested):
@idelcano After merge I'm not sure that this fragment has gone well, can you review it please?

xurxodev (approved):
thanks @idelcano

PR #1572, opened 2017-09-20
Feature push of obs plan

### LINKED ISSUE #1561: Push Obs&Action Plan
AFAIK, pushing the Obs&Action Plan means pushing the whole survey again, because PATCH method is not implemented in DHIS2, and by using PUT we need to resend all the previous DEs in order to avoid squashing the previous data. But, of course, that needs to be verified again. So the task would be to:
- [ ] Verify if a PUT for a few new DEs to a previous event would delete previous data on that event
- [ ] Implement the new logic to manage the queue, to check not only the surveys to be send, but also the Obs&Action plans, and treat them as independent element to be send, just like with surveys
- [ ] Implement Obs&Action plan push, by sending the whole survey again if PUT deletes the previous values for the event, or sending just the new DEs (control DEs) in case PUT doesn't delete previous values

idelcano:
Closes #1561


xurxodev (approved):
thanks @idelcano

PR #1613, opened 2017-10-03
refactor share html action to read data from model and remove duplica…

### LINKED ISSUE #1611: Refactor ShareHtmlText to retrieve data to share from model instead of EditTexts

xurxodev:
close #1611


ifoche (approved):
thanks @xurxodev

PR #1626, opened 2017-10-10
Feature translate only hnqis programs

### LINKED ISSUE #1620: Translate only HNQIS programs on pull
There will be a Program Type Program Attribute on HNQIS programs. Only those that contain Program Type = "HNQIS" will be translated during the pull

idelcano:
Closes #1620


app/src/main/java/org/eyeseetea/malariacare/data/database/datasources/ConversionLocalDataSource.java
```diff
@@ -168,7 +170,15 @@ private void convertMetaData(ConvertFromSDKVisitor converter) {
 
         int count;
         //Dataelements ordered by program.
-        List<ProgramExtended> programs = ProgramExtended.getAllPrograms();
+        List<ProgramExtended> allDhisPrograms = ProgramExtended.getAllPrograms();
```
ifoche: Instead of adding a for loop and including only those programs that match the desired condition (those who are considered as "valid", why don't just create a new DB query and get the proper list that you need?

app/src/main/java/org/eyeseetea/malariacare/data/database/datasources/ConversionLocalDataSource.java
```diff
@@ -362,6 +372,9 @@ private void convertDataValues(ConvertFromSDKVisitor converter) {
             for (ProgramExtended program : ProgramExtended.getExtendedList(
                     SdkQueries.getProgramsForOrganisationUnit(organisationUnit.getId(),
                             ProgramType.WITHOUT_REGISTRATION))) {
+                if(!program.isValidProgram()){
+                    continue;
+                }
```
ifoche: Again, instead of adding this here, I would change the DB query to only retrieve the valid programs

app/src/main/java/org/eyeseetea/malariacare/data/database/iomodules/dhis/importer/ConvertFromSDKVisitor.java
```diff
@@ -584,9 +584,12 @@ public void buildOrgUnitProgramRelation(OrganisationUnitExtended sdkOrganisation
...
                 SdkQueries.getProgramsForOrganisationUnit(appOrgUnit.getUid(),
                         ProgramType.WITHOUT_REGISTRATION))) {
-            sdkProgramExtended.setAppProgram(programMapObjects.get(sdkProgramExtended.getUid()));
+            if(sdkProgramExtended.isValidProgram()) {
+                sdkProgramExtended.setAppProgram(
+                        programMapObjects.get(sdkProgramExtended.getUid()));
 
-            addOrgUnitProgramRelation(sdkOrganisationUnitExtended, sdkProgramExtended);
+                addOrgUnitProgramRelation(sdkOrganisationUnitExtended, sdkProgramExtended);
+            }
```
ifoche: Can we loop here not on all the programs, but only on valid programs. I would change the getProgramsForOrganisationUnit by adding an optional parameter (function overloading) to determine the program type

ifoche (dismissed):
@idelcano I've asked for some changes. Basically they have to do with not looping on what we had before but checking every time if the program is valid, but looping only on valid programs. This is, create new SQL queries to retrieve the data that you want to operate with. Can you take care of it? Thanks in advance!

xurxodev (approved):
thanks @idelcano

Issue #1618, opened 2017-10-09
Receive Program filter from outside and Make monitoring screen accept OU filtering


ifoche:
Now filters is outside html and should receive program filter or org unit filter 
- Modify current logic in web view for receive program filter from outside
- Instead of receiving a program, the js should receive an OU, and behave exactly the same as with a Program

xurxodev:
@idelcano this issue depends on #1515

PR #1631, opened 2017-10-11
Feature abstract filtering from the modules

### LINKED ISSUE #1515: Abstract filtering from the modules
- [ ] Extract filtering layout (dropdowns) outside from the tabs layout to make it independent from modules fragments
- [ ] By default configure ALL option on both filters
- [ ] Save & keep status
- [ ] Add filter selection to Session class 
- [ ] Apply filtering logic. While keeping planning and monitoring exclusive filtering behavior, allow both filtering by OU & Assessment on Assess & Improve tabs 
   - Abstract filters logic to a common single entity with an associated pre-action default to none and overwritten by planning and monitoring modules to apply the fragment change
   - Create links between filters and modules with 2 possible definitions EXCLUSIVE & NON_EXCLUSIVE. Exclusive filtering is for Planning & Monitor
   - For EXCLUSIVE filters on a module, each filter plays a role (master & slave). Slave filter is moved to default option when selected module role is exclusive

xurxodev:
close #1515


xurxodev:
@ifoche filters in monitor not working, @idelcano fix it on issue https://github.com/EyeSeeTea/malariapp/issues/1618

ifoche (approved):
thanks @xurxodev

PR #1638, opened 2017-10-12
Feature receive program and ou filters from java

### LINKED ISSUE #1618: Receive Program filter from outside and Make monitoring screen accept OU filtering

idelcano:
Closes #1618


xurxodev (changes requested):
@idelcano I found two problems:
* when exists a filter previously and I navigate to monitor the tab is empty
* the graphic does not show chart by all org units and all assessments filter

xurxodev (approved):
thanks @idelcano

PR #1639, opened 2017-10-12
Feature make feedback arrow visible only if with not filtered questions

### LINKED ISSUE #1632: On feedback, make expand/collapse arrow only show when there's at least one question inside
Due to the filters to show only failed questions and to show only videos/image questions, it can be sometimes difficult to navigate through the CS to find the questions. I would be much more easy if the arrows are only displayed when there's something to display. This is: For CS containing other CS, it make sense to keep the arrow, but for CS whose questions are not displayed because they're hidden by a filter...the arrow only creates confusion

idelcano:
Closes #1632


xurxodev (changes requested):
@idelcano when a composite score contains another composite score the arrow is hidden if inner composite score has hidden questions by filter but I think only for direct parent in questions where these  questions are hidden by filter should are hidden the arrow

xurxodev (approved):
thanks @idelcano

PR #1642, opened 2017-10-12
Content change plain and formatted text

### LINKED ISSUE #1636: change TXT and HTML text
After alpha testers review, they asked us to make a little change to the text of these options (because users don't understand its meaning).

- TXT must be changed into "Plain text"
- HTML must be changed into "Formatted text"

We need to incorporate this also into 1.2 branch

idelcano:
Closes #1636


xurxodev (approved):
thanks @idelcano

PR #1645, opened 2017-10-13
Feature improve user experience on tab change

### LINKED ISSUE #1643: Improve user experience on tab change
- From assess to improve, if coming from a completed survey, directly show the feedback for that survey
- From improve to monitoring, if a survey was selected, then apply filter for that OU

xurxodev:
close #1643


ifoche (approved):
thanks @xurxodev

PR #1651, opened 2017-10-19
fix show/hide compositescore children with multiple levels

### LINKED ISSUE #1649: Clicking in a cs who is parent of multiple composite score levels doesn't' show or hide the correct childrens - feedbackfragment

idelcano:
Closes #1649 
@xurxodev  you can test this bug with the "ke hnqis imci" program and "3 work enviroment"  composite score and children.


xurxodev (approved):
thanks @idelcano

PR #1682, opened 2017-11-06
Change max monitor events from  constant to max events preference set…

### LINKED ISSUE #1663: variable target on settings

idelcano:
Closes #1663


xurxodev (changes requested):
@idelcano You are used the existed preference Max Loaded surveys and It's not correct because this preference is for pull process, We should create a different and new preference SurveysTarget to draw the target line in monitoring

xurxodev (approved):
thanks @idelcano I have added a max length to preference for avoid a crash

PR #1685, opened 2017-11-06
remove unnecesary survey answered ratio saves

### LINKED ISSUE #1658: saveSurveyAnsweredRatioUseCase is executed many times
saveSurveyAnsweredRatioUseCase is executed many times but if we are saving with every change I think is not necessary

idelcano:
Closes #1658


xurxodev (approved):
thanks @manuelplazaspalacio

PR #1688, opened 2017-11-07
use malaria CustomTextView

### LINKED ISSUE #1671: Review and change views that are not custom
- [ ] Reschedule title
- [ ] Assess headers
- [ ] Action plan spinners

xurxodev:
close #1671


xurxodev:
@ifoche there were custom views that do not adapt according to font changes because we were using the views of the SDK but not the font management that exists in picture, at the moment we can not use the SDK views

ifoche (approved):
thanks @xurxodev

PR #1723, opened 2018-02-08
Showing in monitoring charts only last 6 months surveys.

### LINKED ISSUE #1719: Use the same base for pie chart and table chart in monitoring
pie chart is getting all the assessments while the table is only taking into account the last 6 months. We need to use 6 months for the pie chart

manuelplazaspalacio:
Close #1719


ifoche (approved):
thanks @manuelplazaspalacio

PR #1724, opened 2018-02-09
Getting in surveys schedule date from control date element _Next_asse…

### LINKED ISSUE #1715: Date of next visit is wrong in Obs & Action Plan
Compared with planning module, it's not properly calculated. Looks like a problem in the SQL query that is probably receiving multiple answers but only taking the first one to make the calculation...or something similar.

manuelplazaspalacio:
Close #1715


ifoche (approved):
thanks @manuelplazaspalacio

PR #1737, opened 2018-02-14
Showing message in monitoring if no surveys.

### LINKED ISSUE #1731: Blank screen when no surveys newer than 6 months
See [this screenshot](https://files.slack.com/files-tmb/T8G8SCDGX-F97L8TVQR-33c99ba53b/screenshot_20180213-095826_1024.png)

manuelplazaspalacio:
Close #1731


ifoche (changes requested):
@manuelplazaspalacio with this code, the message is shown instead of the general chart. But now let's imagine that you have surveys for only one program. When you select any of the programs for which you don't have surveys in the last 6 months the problem is still there, and nothing is shown. Could you make it to show also the message in those situations?

ifoche (approved):
thanks @manuelplazaspalacio

PR #1741, opened 2018-02-14
Bug blue background arrow

### LINKED ISSUE #1729: Dropdowns show blue background arrow
Dropdowns still have the old EDS style. Please use a yellow background and move the arrow to the right

manuelplazaspalacio:
Close #1729


ifoche (approved):
Thanks @manuelplazaspalacio. Just for the record, finally this PR is not lacing the arrow at the right end of the dropdown, but changing its background to the same yellow that's used in the rest of the assess module. It will probably need to be slightly modified for QuestMark introduction, but that will be a different war.

PR #1774, opened 2018-02-26
Replacing < with html code.

### LINKED ISSUE #1717: When a < symbol is used on feedback, it doesn't render
Please see [this screenshot](https://user-images.githubusercontent.com/26415203/33883144-bca0ed6e-df4b-11e7-956c-e39e7d87c90f.jpg)

manuelplazaspalacio:
Close #1717


xurxodev (changes requested):
@manuelplazaspalacio we should not convert specific char because is possible that this error occurs with other chars, I think is better if we realize encode to all input texts

manuelplazaspalacio:
In v1.3_hnqis this bug is solved, it puts (Start survey always) put the issue as invalid.

xurxodev (approved):
thanks @manuelplazaspalacio

PR #1775, opened 2018-02-26
Feature crash after share

### LINKED ISSUE #1772: Pressing back after selecting to share the obs&act plan produces unexpected results
Bug for  H1.2 maintennace but happening in H1.3  and not in H1.2

idelcano:
Closes #1772 

(Tested in 4.4.2)
Fixed:
- Crash on null context after resume app.
- Crash on illegalArgumentException formatting completion date.
- Package windows filesystem name problem.
- Null exception when getView is called after onResume 

I will open new issue to find a solution with other related problem, the tab-layout general problem. ->https://github.com/EyeSeeTea/malariapp/issues/1776


ifoche (approved):
Thanks @idelcano. As we said in our conversation, this issue is in reality kind of a minor refactor as it's not apparently solving any crash.

Issue #1752, opened 2018-02-23
Assess module: Don't show + button when not needed


ifoche:
- [ ] If you filter by OU + Program and there's a result visible in the list, you cannot create a new survey because there's already a survey in progress. Remove + button in that situation

PR #1779, opened 2018-02-28
hide plus button when the user can't create new survey with the selec…

idelcano:
…ted filter
Closes https://github.com/EyeSeeTea/malariapp/issues/1752


xurxodev (changes requested):
@idelcano I do not understand why all this new code is necessary (new db queries, etc.) because currently the surveys in progress are being filtered by org unit and program in the list, so I think that if this current filter returns data, then hide the button of a new survey, but a new survey button should be displayed, can you review it, please?

idelcano:
@xurxodev i talked with @ifoche  and i follow your interpretation but checking if the orgunit/program  combination is valid.

xurxodev:
Ok @idelcano but between my comment and your comment there is not any commit, have you forgotten to push changes?

idelcano:
@xurxodev  yes sorry, i pushed in command-line but the push failed and i didn't see the error.

xurxodev (approved):
thanks @idelcano

Issue #1753, opened 2018-02-23
Assess module: Copy filters selection to survey creation dropdowns


ifoche:
If filters are not in "All xxx" option, then copy that selection into the dropdowns ([See image](https://camo.githubusercontent.com/ae0ed2b91e8bf133104c5e48e58deaa17c14f896/68747470733a2f2f776166666c65696f2d6469726563742d75706c6f6164732d70726f64756374696f6e2e73332e616d617a6f6e6177732e636f6d2f75706c6f6164732f3536623336656130633538376465313330306238623536352f313235353136633636653832633732386163653231653064343663306261616230616231633365346366663961643436623735386532383232633036323137613235313061363339333039366536323462613431366630383030303135656562313930643031353461646164646437326638653132633265646631623162616539343266313932313866326630646265663863303430653535346539363439393435393762336261643834386330656536653233653764383836346433653161303631383839666635663363613339613433396234616561353961333163306266303463366233616163323337343665346438363166373232363661363232376138666664613537383138363634616362323134336431622e706e67))

PR #1780, opened 2018-02-28
use filter on create survey if exist

idelcano:
Closes https://github.com/EyeSeeTea/malariapp/issues/1753


xurxodev (approved):
thanks @idelcano

PR #1796, opened 2018-03-07
Feature questmark tabhost

idelcano:
Closes https://github.com/EyeSeeTea/malariapp/issues/1795

Review:
Tabhost cosmetics.


xurxodev (changes requested):
@idelcano I have found some bug:
 - The tab text should change when active status change

xurxodev (approved):
thanks @idelcano

PR #1823, opened 2018-03-19
Feature disable html sharing in obs and action plan

### LINKED ISSUE #1759: Disable HTML sharing in Obs & Action Plan
We need to stop offering 2 options in Obs & Action Plan sharing. So now when clicked it directly should take you to the sharing (showing only one suboption doesn't make sense)

Amejia481:
close #1759


xurxodev (approved):
thanks @Amejia481

PR #1828, opened 2018-03-19
Hiding font customisation option

### LINKED ISSUE #1770: Hide font customisation option

Amejia481:
close #1770


xurxodev (changes requested):
@Amejia481 you have realized this change for all variants, please can you use gradle variable as in PictureApp a realize this change only for HNQIS variant? thanks

Amejia481:
Thanks @xurxodev
You're right. I'll fix it right away!

xurxodev (changes requested):
@Amejia481 It's true that now only HNQIS variant remove font customization settings but you have used strategies. In my previous comment, I suggested you use Gradle variable as in PictureApp because is a more simple solution with an standard functionality.
Look this https://github.com/EyeSeeTea/pictureapp/blob/v0.3_cnm/app/build.gradle#L115
I think is better this approach because only change build.gradle variable value we can to add or remove this feature for a variant.
Can you refactor it? thanks

Amejia481:
Sure!

Amejia481:
Done  @xurxodev!
Thanks for your patience and sorry for all the troubles!

xurxodev (approved):
thanks @Amejia481

PR #1834, opened 2018-03-23
Feature remove filter when moving via bubble click

### LINKED ISSUE #1767: Avoid the application to apply any filter when moving via bubble click from monitoring to improve

idelcano:
Closes #1767


xurxodev (changes requested):
@idelcano on the previous release if I have selected specific surveys filter and all org units filter, when I navigate from monitor to feedback then org unit filters was selected also. I think the issue is: to navigate from monitor to feedback should not modify current filters but your solution resets current filters to all. I have confirmed the issue with @ifoche , can you review it? thanks

xurxodev (approved):
thanks @idelcano

PR #1835, opened 2018-03-23
Feature fix bad answer calc

### LINKED ISSUE #1783: Overall score not reaching 100%
After filling in all the questions from a survey, looks like overall 100% is not reached. 

Please [see this screenshot](https://user-images.githubusercontent.com/26415203/33876956-00edfa3e-df38-11e7-812e-f557d34b071a.jpg) and [also this DB to replicate it](https://github.com/EyeSeeTea/QAApp/files/1558909/compressedData.zip)

idelcano:
Closes #1783 

The problem was the calc of the question type "no-answer" questions.

I fixed it for the future surveys calcs, but if the user has a survey with an active questiontype "no answer", 
the survey call will still wrong for that survey.


xurxodev (approved):
thanks @idelcano

PR #1836, opened 2018-03-23
Added left padding in feedback content only in hnqis

idelcano:
Closes https://github.com/EyeSeeTea/malariapp/issues/1766


xurxodev (approved):
thanks @idelcano

PR #1840, opened 2018-03-29
Solved bug sharing observation & actionPlan

### LINKED ISSUE #1839: Next assessment date is different when shared
Some alpha testers are having the following issue (I've been unable to replicate it):

1) They login
2) Go to a feedback
3) Open the action plan, and mark as complete
4) Share the action plan

If you look at the action plan shared, apparently they get a different next assessment date than the one shown in the app

manuelplazaspalacio:
Close #1839


ifoche (approved):
Thanks @manuelplazaspalacio!! That was not an easy one!

PR #1843, opened 2018-04-02
Feature fix filters workflow problems

### LINKED ISSUE #1786: Filters consistency
Check the consistency among Surveys & OU filters across the modules. It seems that there are weird behaviors when moving from a module to another one in case we have a module still open / not finalized an action (i.e. Assess module still active, Improve module in a feedback or Monitor module with a OU selected).

idelcano:
Closes #1786 
I clean code and the survey services(and fix plannedsurvey service bugs like a crash when you enter in planing with a org unit + program selected).
I did n`t remove the reloadDashboard but i change the reloadData of every fragment from relaodDashboard to reload only the fragment data. I see the app more fast and i can`t replicate some anomalous workflows with the filters after change it.


ifoche (approved):
thanks @idelcano

PR #1858, opened 2018-04-05
Added db table and upgrade version in demo mode

### LINKED ISSUE #1855: Crash when select demo mode
I think we need update the assets database and csv.

Error:
Caused by: android.database.sqlite.SQLiteException: duplicate column name: provider (code 1): , while compiling: ALTER TABLE `ObsActionPlan` ADD COLUMN provider string

idelcano:
Close #1855


xurxodev (approved):
thanks @idelcano

PR #1859, opened 2018-04-05
Disabling share button instead of hiding it on status in progress

### LINKED ISSUE #1763: Share button only show when action Plan is marked as completed
@Amejia481  Share button should not be hidden but disabled and when status is in progress. For status equal to completed or sent should be enabled

Amejia481:
close #1763


xurxodev (approved):
thanks @Amejia481

Issue #1695, opened 2017-11-27
Blank fragment


ifoche:

ifoche:
Example [here](https://camo.githubusercontent.com/d796a6349066a8705be3b0aff20e38a06000a28f/68747470733a2f2f73332d75732d776573742d322e616d617a6f6e6177732e636f6d2f62756464796275696c642d73637265656e73686f74732f3536623430386164363563353637303130306164663464662f3561303262653561343638653939303030313633633865302f33363434626263642d336630642d346336342d396534312d3366313438653133633361352e6a7067)

ifoche:
More clues, they've been able to reproduce it when going to the improve module from sharing an assessment in formatted text

manuelplazaspalacio:
Test in Android 4.4  when you change quickly  between tabs, changing orgunits in spinners, scheduling surveys... The tab bar blocks in a position and puts the rest of the screen in blank.

idelcano:
I continued this issue in #1786

PR #1860, opened 2018-04-05
Feature fix blank layouts

### LINKED ISSUE #1776: Layout disappears
When the user changes the tab very quickly, or when he tries to return from sharing the text, the layouts may disappear.

I think this problem is realted with the SurveyService, the onResume, and the tabcontent workflow.

Suggestion: Try to replicate in 4.4.2

idelcano:
Closes #1776 
I started from #1695 and #1786


ifoche (approved):
thanks @idelcano

PR #1885, opened 2018-04-09
Buf unify dialog aspect

### LINKED ISSUE #1874: Unify dialog menus aspect
- [ ] Margin between 3 dots menu and the list item
- [x] Margins and general aspect for the dialog itself

Note: if we need to change a layout, please copy it on hnqis variant to avoid conflict with eyeseetea variant

manuelplazaspalacio:
Close #1874


xurxodev (approved):
thanks @manuelplazaspalacio

PR #1887, opened 2018-04-09
When a action plan is sent the import summary could be import == 0

### LINKED ISSUE #1868: Double check on Obs&Action plan doesn't appear
St. Consolata Clinic + IMCI

Data is sent to server (the API answers ok) but apparently in the DB it is in SENDING state

idelcano:
Closes #1868
To replicate send a action plan with empty values.


ifoche (changes requested):
@idelcano can you please modify this solution to include also a solution for the problem described in #1896? This solution address the problem of having empty obs&action plan and that other issue the problem of having an empty normal survey. I have the feeling that we can provide a good solution for both at the same time

ifoche (approved):
thanks @idelcano

PR #1910, opened 2018-04-11
Put system defined for current phone language, and add English as lan…

### LINKED ISSUE #1898: If your system default language is not English, once you change the language you cannot come back to the system defined

manuelplazaspalacio:
Close #1898


xurxodev:
Solved in #1911

PR #1919, opened 2018-04-13
Feature create pull demo use case

### LINKED ISSUE #1918: Refactor Pull demo process does not use current for does not use PullUseCase and PullController
Pull process and pull demo process are not the same process, It has different logic and validations (cancel, pull in progress, steps, conversions ...) then we need refactor pull demo process:
- Create PullDemoUseCase
- Rename LocalPullController to PullDemoController, remove current interface implementation to IPullController
- Create a new interface in domain layer for PullDemoController

idelcano:
Closes #1918


xurxodev (approved):
thanks @idelcano

PR #1920, opened 2018-04-14
Feature create import use case

### LINKED ISSUE #1917: Refactor import db process for does not use current PullController
Import DB is a different process of current PullController, then:
- create ImportUseCase and ImportController
- Remember to create an interface in domain layer for ImportController

idelcano:
Closes #1917


xurxodev (approved):
thanks @idelcano. I have refactor a little the code for follow SOLID principles

PR #1922, opened 2018-04-16
Feature split pull controller

### LINKED ISSUE #1899: Split PullController into PullMetadataController & PullDataController
- Please note that this task will imply to move all the logic that can be moved from Controllers to PullUse. For every logic to move, create a separate issue if It does not exists

xurxodev:
close #1899


ifoche (approved):
thanks @xurxodev

PR #1923, opened 2018-04-16
Feature move cancel pull

### LINKED ISSUE #1903: Move cancel pull logic from PullController to PullUseCase
depends on #1899

idelcano:
Closes #1903


xurxodev (changes requested):
@idelcano with this solution the use case decide conversions invoking next step and this is not correct.
Use case should not invoke steps.

 In another meeting we talk about one more simple solution approved by @ifoche . Use case invoke two blocks, pull metadata and pull data  then for this moment only between these two blocks cancellation is possible.

Can you review it? thanks

PR #1926, opened 2018-04-17
Feature move mandatory to use case

### LINKED ISSUE #1908: move mandatory metadata validation from PullController to PullUseCase
depends on #1899

idelcano:
Closes #1908


app/src/main/java/org/eyeseetea/malariacare/domain/usecase/pull/PullUseCase.java
```diff
@@ -63,7 +67,11 @@ private void pullMetadata(final Callback callback) {
 
             @Override
             public void onComplete() {
-                pullData(callback);
+                if(mMetadataValidator.isValid()){
+                    pullData(callback);
+                } else {
+                    onError(new ConversionException());
```
xurxodev: @idelcano Remember that domain should not know anything about conversions then has no sense that the use case creates ConversionException. I think that we should create o refactor to MetadataException where the message can be "All necessary mandatory metadata tables are not populated" for example.
Can you review it? thanks

xurxodev (changes requested):

xurxodev (approved):
thanks @idelcano

PR #1929, opened 2018-04-17
Feature bug blank event

### LINKED ISSUE #1928: empty events bug
This bug causes a exception when the surveyChecker try to find a quarantine survey in the server to set as sent or as completed. 

This is caused by a server response with a empty event. 

The empty event is pushed when the user have a normal survey and a observation and plan action survey. 

The observation and plan action survey push wipe the sdk database during the normal survey push, and the events will pushed by the sdk without datavalues because the datavalues was setted in the middle of the push after get the empty events.

idelcano:
Closes #1928


ifoche (approved):
thanks @idelcano

PR #1944, opened 2018-04-19
Added click dialog on chart with no complete compulsory question surveys

### LINKED ISSUE #1935: Give access to completion dialog also when it's not 100% completed
When you click on the top-right double pie chart, only if you're 100% mandatory completed you can see the dialog to complete the survey. The client wants to add a new functionality, so the dialog is also accessible when you're not 100% complete and the complete button is greyed out, to avoid the user completing the survey when it's not allowed

idelcano:
Closes #1935


xurxodev (approved):
thanks @idelcano

PR #1945, opened 2018-04-19
Allow reschedule surveys with the default date

### LINKED ISSUE #1939: On massive rescheduling, reschedule button must always be active, and not only when you change the date
some selected surveys may not have the same scheduled date

idelcano:
Closes #1939


xurxodev (approved):
thanks @idelcano

PR #1948, opened 2018-04-20
fix schedule strings

### LINKED ISSUE #1942: Massive rescheduling still uses some hardcoded strings instead of POEditor ones
If you change to French you'll see the titles "Schedule" and "Due date" there, but they exist in the strings. We need to start using them

idelcano:
Closes #1942


xurxodev (approved):
thanks @idelcano

PR #1967, opened 2018-05-08
Feature use executors from push use case

### LINKED ISSUE #1965: Use executors (main and async) from the PushUseCase as in PictureApp

xurxodev:
close #1965


ifoche (approved):
thanks @xurxodev!

PR #1985, opened 2018-05-23
Feature questmark remove is cached visibility

### LINKED ISSUE #1983: Remove isCachedVisibility from Question domain Entity
isCachedVisibility is not a domain concept, It's a AnswerRatio calculation concept,  refactor Question for remove it and implement solution in another way.

idelcano:
Closes #1983


xurxodev (approved):
thanks @idelcano. I have moved QuestionType from data layer to domain layer

PR #1988, opened 2018-05-24
Feature create server metadata entity

### LINKED ISSUE #1982: Create ServerMetadata domain entity
Create domain entity with best practices and add all necessary code only for being used for the pull, some tips:

* Avoid public setters by default, please read this article
http://xurxodev.com/clean-architecture-code-smells-parte-2/
* create a constructor with mandatory fields
* add relations to other entities by uid (no object reference)
* add validations to mandatory info when a new instance is created
* create unit tests
* create the repository for retrieve ServerMetadata by program uid

!!! Important, this class have not structure as database, this class should have a getter by control data element, CDE codes  is in donotranslate file:
For the moment only are necesarry this CDE:
    private final String nextAssessmentUid;
    private final String creationDateUid;
    private final String completionDateUid;
    private final String uploadDateUid;
    private final String uploadByUid;
We should realize a query to database and convert to CDE uid from CDE code in donotranslate

idelcano:
Closes #1982


xurxodev (approved):
thanks @idelcano

PR #2001, opened 2018-05-31
fix calendar on click listener ignored when no focus problem

### LINKED ISSUE #2000: When rescheduling, user need to double click on the calendar make it show
First click gets the focus and second one opens the calendar. We need that it opens with the first click

idelcano:
Closes #2000


xurxodev (approved):
thanks @idelcano

PR #2004, opened 2018-06-04
ignore options without code during conversion from sdk to dhis

### LINKED ISSUE #1996: Bug retrieving Options from repository
There are options in db with code null and this field is required in domain

idelcano:
Closes #1996


xurxodev (approved):
thanks @idelcano

PR #2012, opened 2018-06-20
Add server URL query string parameter

### LINKED ISSUE #2009: Add server URL query string parameter
Add server URL parameter so it becomes available for all servers

https://apps.psi-mis.org/hnqis/feedback?eventId=iWgYrkhpdlh&server=https://clone.psi-mis.org

Use the same setting that is already in place "DHIS2 server URL"

manuelplazaspalacio:
Close #2009

### DIFF
--- a/app/src/hnqis/res/values/donottranslate.xml
+++ b/app/src/hnqis/res/values/donottranslate.xml
@@ -29,5 +29,5 @@
 
     <!-- pull control codes -->
     <string name="pull_program_code" translatable="false">HNQIS</string>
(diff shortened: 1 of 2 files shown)


xurxodev (changes requested):
@manuelplazaspalacio, 
according to how I understand the description of the issue by @ifoche:
**Add server URL parameter so it becomes available for all servers.
https://apps.psi-mis.org/hnqis/feedback?eventId=iWgYrkhpdlh&server=https://clone.psi-mis.org
Use the same setting that is already in place "DHIS2 server URL"**

Server URL parameter should be dynamic reading from DHIS2 server URL

xurxodev (approved):
thanks @manuelplazaspalacio

PR #2024, opened 2018-06-26
Feature fix user in session

### LINKED ISSUE #2022: Sometimes user push CDE upload with different values for the same server user
Control data element upload with different values for the same user (name, username, uid)

- for this event, https://data.psi-mis.org/api/events/Lm9C3K4BV5n upload data value user push equal to bWg48MMAQhR (uid)

- for this event, https://data.psi-mis.org/api/events/NpAPthSXtL6 upload data value user push equal to KEdemo1 (username)

- for this event, https://data.psi-mis.org/api/events/gFqMeJWxym7 upload data value user push equal to Kenya Demo1 (first name and last name)

On pull process, we find the user in the database by data value in CDE and if not exists create one new then after pull process there are 3 or 4 user but related user server is the same.

idelcano:
Closes #2022 

i could not replicate the problem but i solved it using always the  username instead of the name
This fix other bug, because the client wants the username instead of the name.

### DIFF
--- a/app/src/main/java/org/eyeseetea/malariacare/DashboardActivity.java
+++ b/app/src/main/java/org/eyeseetea/malariacare/DashboardActivity.java
@@ -70,9 +70,6 @@ public class DashboardActivity extends BaseActivity {
             new AsyncAnnouncement().execute();
         }
 
-        //XXX to remove?
-        initDataIfRequired();
-
         loadPhoneMetadata();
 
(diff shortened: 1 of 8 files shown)


xurxodev (changes requested):
@idelcano with this solution app crash after pull process because currently in development after pull process there is not any user with a username.

xurxodev (approved):
thanks @idelcano

PR #2028, opened 2018-06-29
Feature extract user remote data source

### LINKED ISSUE #2017: Extract user annoucement and closed date calls to a IUserRemoteDataSource
Create a IUserRemoteDataSource to get the user annoucement and closed date
Use a User domain entity with annoucement and closed date fields.

Extract user calls, call from PullDhisApiDataSource to IUserRemoteDataSource.
depends on #2018

idelcano:
Closes #2017 
@xurxodev i created a date parser class from eventExtendend. The intent of this class is remove all common date parser methods from eventExtended.
And i created a use case to pull and treat the user attributes.


app/src/main/java/org/eyeseetea/malariacare/domain/entity/UserAttributes.java
```diff
@@ -0,0 +1,68 @@
...
+package org.eyeseetea.malariacare.domain.entity;
+
+
+import org.eyeseetea.malariacare.utils.DateParser;
+
+import java.util.Date;
+
+import static org.eyeseetea.malariacare.domain.utils.RequiredChecker.required;
+
+public class UserAttributes {
```
xurxodev: This class has no sense in our domain because save these user concepts as attributes are dhis domain concepts, our domain has no attributes. This class represents user Announcement message and IsClose properties of UserAccount in our domain then we should create this getters in UserAccount directly.

app/src/main/java/org/eyeseetea/malariacare/domain/usecase/PullLoggedUserAttributesUseCase.java
```diff
@@ -0,0 +1,98 @@
...
+import org.eyeseetea.malariacare.data.database.utils.PreferencesState;
+import org.eyeseetea.malariacare.domain.boundary.IUserAttributesRemoteDataSource;
+import org.eyeseetea.malariacare.domain.boundary.IUserLocalDataSource;
+import org.eyeseetea.malariacare.domain.boundary.executors.IAsyncExecutor;
+import org.eyeseetea.malariacare.domain.boundary.executors.IMainExecutor;
+import org.eyeseetea.malariacare.domain.entity.UserAccount;
+import org.eyeseetea.malariacare.domain.entity.UserAttributes;
+import org.eyeseetea.malariacare.domain.exception.PullUserAttributesException;
+
+public class PullLoggedUserAttributesUseCase implements UseCase {
```
xurxodev: This use case has no sense, we should create GetUserAccountUseCase

app/src/main/java/org/eyeseetea/malariacare/domain/usecase/PullLoggedUserAttributesUseCase.java
```diff
@@ -0,0 +1,98 @@
...
+
+    public interface Callback {
+        void onSuccess(UserAttributes userAttributes);
+
+        void onError(UserAttributes userAttributes);
+    }
+
+    private final IAsyncExecutor mAsyncExecutor;
+    private final IMainExecutor mMainExecutor;
+    private final IUserLocalDataSource userLocalDataSource;
```
xurxodev: A use case should be agnostic with data, It should execute UserAccountRepository and this to retrieve data from the local data source or remote data. The logic to execute remote or local can be implemented for several ways, one can be with time if the user data stored in local has more of 1 days for example then return from the remote. Sometimes this logic can be realized by a parameter to use case and repository like NetworkStrategy (NetworkFirst and LocalFirst). I think that Current UserRepositoy (with login and logout) should be rename to AuthenticationManager like in PictureApp and create new UserAccountRepository to retrieve UserAccount from local or remote according to needs

app/src/main/java/org/eyeseetea/malariacare/domain/boundary/IUserLocalDataSource.java
```diff
@@ -0,0 +1,9 @@
+package org.eyeseetea.malariacare.domain.boundary;
+
+import org.eyeseetea.malariacare.domain.entity.UserAccount;
+
+public interface IUserLocalDataSource {
```
xurxodev: Abstractions should not be local or remote, only implementations. This interface should be IUserAccountRepository

app/src/main/java/org/eyeseetea/malariacare/data/remote/api/UserAttributesRemoteDataSource.java
```diff
@@ -0,0 +1,81 @@
...
+
+import com.fasterxml.jackson.databind.JsonNode;
+
+import org.eyeseetea.malariacare.data.database.model.UserDB;
+import org.eyeseetea.malariacare.domain.boundary.IUserAttributesRemoteDataSource;
+import org.eyeseetea.malariacare.domain.entity.Credentials;
+import org.eyeseetea.malariacare.domain.entity.UserAttributes;
+import org.eyeseetea.malariacare.domain.exception.PullUserAttributesException;
+
+public class UserAttributesRemoteDataSource extends OkHttpClientDataSource implements IUserAttributesRemoteDataSource {
```
xurxodev: Logic in this class should be in a class like:
public class UserAccountAPIDataSource implements IUserAccountDataSource

app/src/main/java/org/eyeseetea/malariacare/domain/boundary/IUserAttributesRemoteDataSource.java
```diff
@@ -0,0 +1,8 @@
+package org.eyeseetea.malariacare.domain.boundary;
+
+import org.eyeseetea.malariacare.domain.entity.UserAttributes;
+import org.eyeseetea.malariacare.domain.exception.PullUserAttributesException;
+
+public interface IUserAttributesRemoteDataSource {
```
xurxodev: this interface has no sense because UserAttributes should not be an entity

xurxodev (changes requested):
@idelcano I have found several problems in this PR.
we should not bring dhis domain concepts to our domain, attributes is an Dhis concept, it is not an EyeSeeTea concept. Please review Clean Architecture concepts again because there are concepts bad applied.
I have added comments in code and basically, we need to apply the diagram we have talked about many times already.
clean architecture

idelcano:
@xurxodev done. 
I created a user in android server KEdemo2 to test this features.
You can check the announcement, and closed date. 
The closed date is  31/07/2018 to test it. 
If you want try to logging with a closed user. you will need update that date in the server (Users/ KEdemo2 / edit user), if you trying increasing the android system date you will get exception related with certificate expiration date .

app/src/main/java/org/eyeseetea/malariacare/domain/usecase/GetUserAccountUseCase.java
```diff
@@ -0,0 +1,93 @@
...
+
+    public void execute(final Callback callback) {
+        this.mCallback = callback;
+        mAsyncExecutor.run(this);
+    }
+
+    @Override
+    public void run() {
+        //get logged user
+        UserFilter userFilter = new UserFilter("", true);
```
xurxodev: In this use case there are some improvements to realize:
- An UseCase should not invoke to  PreferencesState.getInstance().setUserAccept
- UserAccept is a concept of the user then we should be in UserAccount and in database
- The logic to set UserAccept to false should be inside UserAccount entity because is a Bussiness Rule logic and the values to decide are inside UserAccount and we should create a unit test to validate the logic
- An UseCase should not execute the repository first with LocalFirst and then with NetworkFirst
- UserFilter should be passed from outside to UseCase

app/src/main/java/org/eyeseetea/malariacare/data/repositories/UserAccountRepository.java
```diff
@@ -1,125 +1,40 @@
...
+    public UserAccountRepository(IUserAccountDataSource userAccountAPIDataSource,
+                                 IUserAccountDataSource userAccountLocalDataSource){
+        this.userAccountAPIDataSource = userAccountAPIDataSource;
+        this.userAccountLocalDataSource = userAccountLocalDataSource;
     }
 
     @Override
-    public void logout(final IRepositoryCallback<Void> callback) {
+    public UserAccount getUser(UserFilter userFilter, NetworkStrategy networkStrategy) {
+        UserAccount userAccount = userAccountLocalDataSource.getUser(userFilter);
```
xurxodev: if NetworkStrategy is NetworkFirst always is invoked userAccountLocalDataSource, this one only should be executed if NetworkStrategy is LocalFirst or if NetworkFirst call fail

app/src/main/java/org/eyeseetea/malariacare/data/NetworkStrategy.java
```diff
@@ -0,0 +1,5 @@
+package org.eyeseetea.malariacare.data;
+
+public enum  NetworkStrategy {
```
xurxodev: NetworkStrategy should be in the domain layer. Now this creates a dependency with the data layer from UserAccountUseCase

app/src/main/java/org/eyeseetea/malariacare/data/remote/api/UserAccountAPIDataSource.java
```diff
@@ -0,0 +1,129 @@
...
+                        getUserAnnouncement(jsonNodeArray, newMessage, i,
+                                UserDB.ATTRIBUTE_USER_ANNOUNCEMENT);
+                closeDate = getUserCloseDate(jsonNodeArray, closeDate, i);
+            }
+            announcement =newMessage;
+            closedDateAsString=closeDate;
+            username = getUsername(jsonNode);
+            name = getName(jsonNode);
+        } catch (Exception ex) {
+            Log.e(TAG, "Cannot read user last updated from server with");
```
xurxodev, line 69: We should not return null UserAccount to avoid work with nulls, we should throw the exception and if in network first throw an exception then the repository should return local data

xurxodev (changes requested):
@idelcano now this PR it looks better but there are still some problems to solve. I have inserted comments in the code.
I think that we should review this PR together in Pair Programming session to avoid to spend a lot of time in questmark issues according to meeting with @ifoche the last week.

xurxodev (approved):
thanks @idelcano

PR #2043, opened 2018-07-09
Feature use composite score domain entity in pull data

### LINKED ISSUE #2003: Use in SurveysDhisDataSource and Surveymapper CompositeScore domain entity and repository
- Currently in SurveysDhisDataSource and SurveyMapper is using CompositeScoreDB, It should use domain entity and repository as for question, option etc..

xurxodev:
### :pushpin: References
* **Issue:** close #2003 
* **Related pull-requests:**

The goal is to use the composite domain entity in SurveyMapper to avoid a dependency from domain to the data layer. Currently, the dependency is with CompositeScoreDB.

Replace from CompositeScoreDB to CompositeScore and modify Repository to retrieve all CompositeScores. I have updated uni tests

 **Use case 1:** Execute uni tests.
 **Use case 2:** Realize a pull in HNQIS and the app should work.

### :art: UI changes?


ifoche (approved):
@xurxodev tests are not working for me. Can you have a look to this? Specially SurveyMapperShould that's added here is not passing

ifoche (changes requested):
actually I said "Approved" but it was a "Request changes" :)

xurxodev:
@ifoche only for the record, seems an android studio bug executing test from project panel, executing ./gradlew data:testEyeseeteaDebugUnitTest  and ./gradlew data:testHnqisDebugUnitTest the test working like in Travis

ifoche (approved):
thanks @xurxodev. Great that you found a way to upgrade gradle later, because this bug is really annoying. All fine, approving.

PR #2084, opened 2018-08-07
Feature manage not supported server version error from sdk

### LINKED ISSUE #2074: Update sdk /api/me change in 2.30 and as a consequence the programs are not downloaded on that call
SDK branch to work 2.30_EyeSeeTea

xurxodev:
### :pushpin: References
* **Issue:** close #2074
* **Related pull-requests:** https://github.com/EyeSeeTea/dhis2-android-sdk/pull/402

Now the 2.30_EyeSeeTea SDK throw an UnsupportedVersionException in login process if the dhis server version is smaller than 2.26 then the goal of this PR is manage the error and show a message to the user.

I have added a new method to LoginUseCase.Callback and create a new string to show the message.

- [x] **Use case 1:** Realize login against https://data.psi-mis.org (2.25), then login should fail and show unsupported server version message.

### :art: UI changes?

unsupported_server_version


ifoche (approved):
thanks @xurxodev

PR #2093, opened 2018-08-07
Feature dorpdown amount data

### LINKED ISSUE #2079: Include a dropdown to select the amount of data we want to download
We already did this in one of the malaria case surveillance. We would need to import from there the same solution with no data by default

manuelplazaspalacio:
### :pushpin: References
* **Issue:** close #2079 
* **Related pull-requests:** 

Add a date filter in the login activity, to select the amount of data you want to download.

I add a spinner in the login activity, and  saving the selection in the shared preferences, then getting from it when starting the pull and configure the pull whit the saved values.

- [x] **Use case 1:** Test login and pull
    - [x] make login in http://eds.jhpiegodb.org:8080 with ifoche_test.
    - [x] check the log, to check if the request are well made it (the start date and the end date filters are the correct).
    - [x] See the correct surveys downloaded in the app.

### :art: UI changes?

### DIFF
--- a/app/src/main/java/org/eyeseetea/malariacare/database/iomodules/dhis/importer/PullController.java
+++ b/app/src/main/java/org/eyeseetea/malariacare/database/iomodules/dhis/importer/PullController.java
@@ -22,8 +22,6 @@ package org.eyeseetea.malariacare.database.iomodules.dhis.importer;
 import android.content.Context;
 import android.util.Log;
 
-import com.raizlabs.android.dbflow.runtime.transaction.process.ProcessModelInfo;
-import com.raizlabs.android.dbflow.runtime.transaction.process.SaveModelTransaction;
 import com.raizlabs.android.dbflow.sql.language.Select;
 import com.squareup.otto.Subscribe;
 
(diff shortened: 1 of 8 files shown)


app/src/main/java/org/eyeseetea/malariacare/database/iomodules/dhis/importer/PullController.java
```diff
@@ -170,12 +169,7 @@ public void pull(Context ctx) {
             //Pull new metadata
             postProgress(context.getString(R.string.progress_pull_downloading));
             try {
-                if(AppSettingsBuilder.isDownloadOnlyLastEvents()){
```
xurxodev, line 173: the isDownloadOnlyLastEvents functionality should not be deleted, we should download the last event filtering by dates

xurxodev (changes requested):
@manuelplazaspalacio I have found some problems.
- You have removed one old functionality, I have added comments in the code.
- When no data is selected, the app should not realize any request to download events, currently realize request but with future dates

xurxodev (approved):
thanks @manuelplazaspalacio

PR #2095, opened 2018-08-08
No reloading the active module if the open fragment is a FeedbackFrag…

### LINKED ISSUE #2085: Improve: back arrows don't take the user to the right place after watching videos
When the user clicks on the back arrow after watching a video, the device should take the user to the Feedback page and not the Improve module landing page.

manuelplazaspalacio:
### :pushpin: References
* **Issue:** close #2085  
* **Related pull-requests:** 

Not closing the Feedback of the showing survey when pressing back in the video activity.

Not reloading module if FeedbackFragment is open.

    - [x] make login in http://clone.psi-mis.or with KEdemo1.
    - [x] got to feedback and select the program KE HNQIS Family Planning Long-...
    - [x] View a survey feedback and open a video and them press back.

### :art: UI changes?


xurxodev (changes requested):
@manuelplazaspalacio This solution work but You have realized this implementation in Dashboardcontroller that It's generic to all modules. Can you realize the implementation in ImproveModuleController that is specific to improve (feedback, sent and action plan)?

xurxodev (approved):
thanks @manuelplazaspalacio

PR #2097, opened 2018-08-08
Change the message in case inactive selection of pair OU/Health Area

### LINKED ISSUE #2087: Assess: different message in case of inactive selection of pair OU/Health Area
In case the pair OU/Health Area is not active because that Health Area is not assigned to the selected OU, if possible, let’s display a different message: ’This survey is not assigned to this facility. Please, select another survey.’
The current message doesn't make sense because the user can't start any new survey.

[See image](https://user-images.githubusercontent.com/12784825/43081237-3c913972-8e9a-11e8-965a-4749a68e6e7c.png)

manuelplazaspalacio:
### :pushpin: References
* **Issue:** close #2087   
* **Related pull-requests:** 

Change the showing message if pair OU/Health is inactive.

Changing the showing message when the add button is hidden.

    - [x] make login in http://clone.psi-mis.or with KEdemo1.
    - [x] got to new surveys tab an select KE HNQIS HIV  and _Test KE 3to see the original message.
    - [x] then change KE HNQIS HIV to KE HNQIS Hypertension to see the new one.

### :art: UI changes?

### DIFF
--- a/app/src/hnqis/res/values-es/strings.xml
+++ b/app/src/hnqis/res/values-es/strings.xml
@@ -288,4 +288,5 @@ clínica"</string>
 <string name="improve_no_surveys">"No se completaron encuestas para la selección que ha realizado."</string>
 <string name="scheduled">Programado</string>
 <string name="reschedule_title_multiple_survey">"%1$d número de encuestas que se reprogramarán para la OU %2$s"</string>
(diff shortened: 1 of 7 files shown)


xurxodev (approved):
thanks @manuelplazaspalacio

PR #2098, opened 2018-08-08
Showing gray background when 0 score in Feedback

### LINKED ISSUE #2089: Show grey background for 0 scored in Feedback
[See image](https://user-images.githubusercontent.com/12784825/43081802-21cbb8e6-8e9b-11e8-9295-9c1d1d86c4c0.png)

Instead of red background, for those that are 0 scored, they want the background to be the same grey as with NA responses

manuelplazaspalacio:
### :pushpin: References
* **Issue:** close #2089   
* **Related pull-requests:** 

Instead of red background, for those that are 0 scored, they want the background to be the same grey as with NA responses.

Adding a new score type for 0 scored and change the color when is scored that type to gray.

    - [ ] make login in http://clone.psi-mis.or with KEdemo1.
    - [ ] got to new survey and select the program KE HNQIS HIV and create a new survey.
    - [ ] press back button and complete it not filling anything.
    - [ ] got to feedback and see the survey feedback.

### :art: UI changes?

device-2018-08-08-165657


app/src/main/java/org/eyeseetea/malariacare/domain/entity/ScoreType.java
```diff
@@ -24,14 +24,15 @@
 
 public class ScoreType {
 
-    public enum Classification {HIGH, MEDIUM, LOW}
+    public enum Classification {HIGH, MEDIUM, LOW, NO_RESPONSE}
```
xurxodev: NO_RESPONSE is an incorrect name because score 0 is when the score calculation result is 0 and this is possible with no responses or negative responses. A better name would be NO_SCORE

app/src/main/java/org/eyeseetea/malariacare/layout/adapters/survey/FeedbackAdapter.java
```diff
@@ -202,6 +202,8 @@ private View getViewByCompositeScoreFeedback(CompositeScoreFeedback feedback, Vi
                 color = R.color.medium_score_color;
             }else if(scoreType.getClassification() == ScoreType.Classification.HIGH) {
                 color = R.color.high_score_color;
+            } else if (scoreType.getClassification() == ScoreType.Classification.NO_RESPONSE) {
+                color = R.color.textFieldColor;
```
xurxodev: Exists a color palette for score, please create a new color in as no_score_color and its values be textfieldcolor

xurxodev (changes requested):
@manuelplazaspalacio  I have found some problems with this PR.

- I have added some problems in code
- There is one failed test in ScoreTypeShould 
- A ScoreType should have a string type because this type is uploaded as control data element to the server, with your solution getType return empty and this is wrong
- Please talk with @ifoche to know if 0 score is type C or a new type
- Please add new tests for new functionality with 0 score

xurxodev (approved):
thanks @manuelplazaspalacio

PR #2104, opened 2018-08-15
Feature dropdown ou tree

### LINKED ISSUE #2080: Include a dropdown to select if you want to download the OU tree
Just like with the dropdown to select the amount of data, we need another dropdown reading "donwload the full OU tree?" with no by default. This should avoid all the OU tree download and build process during the pull. 

As a subtask here, we will need to change the survey creation screen to show only 1 dropdown for the OU whenever there's no OU tree downloaded.

manuelplazaspalacio:
### :pushpin: References
* **Issue:** close #2080   
* **Related pull-requests:** 

Show a spinner to select to download all OU tree or not.

Adding the spinner to select it, save it in preferences and  getting it when downloading OU.

- [ ] **Use case 1:** Open the app, press on text "show advanced options" , select to not download OU tree and do login.
- [ ] **Use case 2:** Open the app, press on text "show advanced options" , select to download OU tree and do login.

### :art: UI changes?

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -72,6 +72,7 @@ android {
             minSdkVersion 15
             targetSdkVersion 23
             buildConfigField "boolean", "loginDataDownloadPeriod", "true"
+            buildConfigField "boolean", "loginOrgUnitTreeDownload", "true"
         }
         hnqis {
             applicationId "org.eyeseetea.malariacare.hnqis_bb"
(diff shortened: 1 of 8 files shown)


xurxodev (approved):
thanks @manuelplazaspalacio 
I have added an empty option to the dropdown as we have already spoken, you forgot it

PR #2117, opened 2018-09-06
Feature implement get surveys method in surveylocaldatasource

### LINKED ISSUE #2044: uncouple push: implement getSurveys method in SurveyLocalDataSource to use from PushController
Require #1904
Requires #2065

xurxodev:
### :pushpin: References
* **Issue:** close #2044 
* **Related pull-requests:**

SurveyLocalDataSource was created on pull issues where dhis2 was uncoupled, The getSurveys method to retrieve Completed surveys was empty and the goal of this issue is to implement getSurveys method that It will use on PushController in next issues. Also on next issues, this method will be refactored to retrieve Quarantine survey using SurveyFilter.

There is one SurveyFilter class that contains all possible filters to retrieve surveys, then this method to use DBFlow queries to retrieve surveys from DB according to filter and use SurveyMapper to convert from surveysDB to domain surveys

 **Use case 1:** How we cannot create tests with dbflow in-memory database in malaria project, then Add this code in push controller under line 74 in PushController class:

            SurveyFilter surveyFilter = SurveyFilter.Builder.create()
                    .WithSurveysToRetrieve(SurveyFilter.SurveysToRetrieve.COMPLETED)
                    .build();

            List<Survey> domainSurveys = new SurveyLocalDataSource().getSurveys(surveyFilter);

And verify that this news code returns the same surveys that old code in line 74 but the difference is that new code return domain surveys.

### :art: UI changes?


ifoche (approved):
thanks @xurxodev !

PR #2130, opened 2018-09-19
pass survey as parameter instead of get from Session

### LINKED ISSUE #2129: Bug saving survey in assess
Create two surveys, move to assess, and complete one. When you complete the second, the survey used to complete is the first all the attempts.

idelcano:
### :pushpin: References
* **Issue:** close #2129
* **Related pull-requests:** 

Set the correct survey to be used by  "mark as complete" action on assess.

Use the active survey and pass it as param instead of get it from session

The problem was in the survey used in "mark as complete" dialog, it was getting the survey in session (the last survey used in SurveyFragment) instead of the active survey.

- [ ] **Use case 1:** Create multiple surveys. Mark as complete all of them in assess tab checking the completed survey was the selected (for example, openning feedback after mark as complete).

### :art: UI changes?


xurxodev (approved):
thanks @idelcano

PR #2139, opened 2018-09-28
Feature add nextschedulematrix

### LINKED ISSUE #2134: Introduce nextScheduleMatrix
- [ ] Instead of using hardcoded variables create an object returning the values. The object would obtain that value from a hardcoded matrix (List of Lists). When we ask for a value, we provide the server we are connected to. If the server doesn't exist, the default matrix is used (so the current values). Otherwise, then the appropriate value is provided.

The matrix would look like this:
Any server : +2, +4 +6 
https://zw.hnqis.org/ : +1, +1, +6

idelcano:
### :pushpin: References
* **Issue:** close #2134  

Introduce nextScheduleMatrix

 Instead of using hardcoded variables create an object returning the values. The object would obtain that value from a hardcoded matrix (List of Lists). When we ask for a value, we provide the server we are connected to. If the server doesn't exist, the default matrix is used (so the current values). Otherwise, then the appropriate value is provided.

The matrix would look like this:
Any server : +2, +4 +6
https://zw.hnqis.org/ : +1, +1, +6

I created a server entity and replaced all the getServerUrl calls to getServer().getUrl() in preferenceState.
The hardcoded matrix is in PreferenceState and when the NextScheduleMonths is created pass the correct array.
I created test to test it too.

- [ ] **Use Case 1:**  complete a survey and the survey in planning should be the correct months

### :art: UI changes?

### DIFF
--- a/app/src/main/java/org/eyeseetea/malariacare/data/database/utils/PreferencesState.java
+++ b/app/src/main/java/org/eyeseetea/malariacare/data/database/utils/PreferencesState.java
@@ -33,6 +33,8 @@ import org.eyeseetea.malariacare.DashboardActivity;
 import org.eyeseetea.malariacare.ProgressActivity;
 import org.eyeseetea.malariacare.R;
 import org.eyeseetea.malariacare.domain.entity.Credentials;
+import org.eyeseetea.malariacare.domain.entity.NextScheduleMonths;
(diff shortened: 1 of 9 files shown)


xurxodev (approved):
thanks @idelcano

PR #2141, opened 2018-10-01
Feature add login dropdown v1.3

### LINKED ISSUE #2136: Add a dropdown in login screen to select the server
Server would change into a dropdown menu with the following options:
- data.psi-mis.org <-- Default selected
- sfhnigeria.psi-mis.org
- singapore-dev.psi-mis.org
- zw.hnqis.org
- Other

When other is selected a textfield appears to write the server address.

idelcano:
### :pushpin: References
* **Issue:** close #2136  

Server would change into a dropdown menu with the following options:

data.psi-mis.org <-- Default selected
sfhnigeria.psi-mis.org
singapore-dev.psi-mis.org
zw.hnqis.org
Other
When other is selected a textfield appears to write the server address.

I added a copy of the dhis login layout and i added the spinner and some IDs.
When a item is selected in the spinner the serverEditText sets his value.
When the user select the "other" option the serverEditText will be shown.

Note: change the sdk to 2.25 !!!

- [ ] **Use Case 1:**  Install app and select a server. The app should login in the selected server.
- [ ] **Use Case 1:**  Install app and select other option. The app should display the edittext, and use its value to use in login.

### :art: UI changes?

newspinner

newspinner2

### DIFF
--- a/app/src/eyeseetea/res/values-es/strings.xml
+++ b/app/src/eyeseetea/res/values-es/strings.xml
@@ -273,4 +273,6 @@ clínica"</string>
 <string name="improve_no_surveys">"No se completaron encuestas para la selección que ha realizado."</string>
 <string name="scheduled">Programado</string>
 <string name="reschedule_title_multiple_survey">"%1$d número de encuestas que se reprogramarán para la OU %2$s"</string>
+
+<string name="other">Otro</string>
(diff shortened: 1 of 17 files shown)


xurxodev (approved):
thanks @idelcano

PR #2144, opened 2018-10-02
Feature update gradle libs

### LINKED ISSUE #2132: Update libraries on build.gradle
We would need to update all libraries versions to the last version possible without needing to touch the code (or at least not too much). Also changing the gradle version when possible would be desirable.

manuelplazaspalacio:
### :pushpin: References
* **Issue:** close #2132  

Update the libs in gradle file.

Changing the version of the libs in gradle an somo needed changes y ActionBarActivities

Run the app and check that everything works correctly

### :art: UI changes?


xurxodev (changes requested):
@manuelplazaspalacio please review the app functionality too because there are metadata and data download that does not work on this branch with libraries upgrade.
After pull, org units spinner has no data and events does not exists in improve (sent) list.
The app is not operative on this branch, testing on 1.4 branch the app work successfully, so the problems is related to upgrade of libraries

xurxodev (approved):
thanks @manuelplazaspalacio

PR #2153, opened 2018-10-10
add new message on feedback share if the survey is not sent

### LINKED ISSUE #2150: Change message when URL is not yet ready for sharing the feedback
When you try to share a feedback that has not yet been sent to the server, the message is not useful. Change it by this one:

'The feedback URL is not available because the survey hasn't been sent to the server yet. Try sharing the feedback again later.'

Please take this change also into 1.4

idelcano:
### :pushpin: References
* **Issue:** close #2150  
* **Related pull-requests:**

Change message when URL is not yet ready for sharing the feedback

Added presenter method to show the alternative feedback text when the survey is not sent

 **Use case 1:** Open a sent survey feedback, open plan action click on share and see the shared text
 **Use case 1:** Open a not sent survey feedback, open plan action click on share and see the new text in shared text

### :art: UI changes?


xurxodev (approved):
thanks @idelcano

Issue #2171, opened 2018-10-25
Review resources in questmark


xurxodev:
Eyeseetea resources should be in main and only variation resources in hnqis variant, avoid duplicate resources

PR #2176, opened 2018-11-02
Feature move eyeseetea resources to main

idelcano:
### :pushpin: References
* **Issue:** connect #2171 

Eyeseetea resources should be in main and only variation resources in hnqis variant, avoid duplicate resources

Move the eyeseetea and common resources to main. And fix some missing resource.

 **Use case 1:** Open hnqis
 **Use case 1:** Open eyeseetea

### :art: UI changes?


xurxodev (changes requested):
@idelcano This issue contains two parts according to description:
1 - Move all eyeseetea resources to main - It's OK
2 - Only variation resources should be in hnqis variant - It is not OK
There are many layouts duplicate in hnqis variant and main:
For example, activity_progress.xml is equal but only change background and icon, this changes can be solved using styles and themes.
Please review resources in HNQIS variant to avoid duplicates.

xurxodev (approved):
thanks @idelcano

PR #2180, opened 2018-11-07
fix demo database missing codes and up to version 17

### LINKED ISSUE #2168: crash when enter in demo mode->logout->normal login
I think this crash is related with the appdatabase version

### LINKED ISSUE #2179: Crash when enter some observation plan´s value in demo mode

idelcano:
### :pushpin: References
* **Issue:** close  #2179   
* **Issue:** close  #2168 (this problem is solved here too)

Fix crash when set some value in a observation field

Updated the demo database to version 17 adding all the new tables and fields and fixing some server metadata codes

 **Use case 1:** Start app from 0 and login in demo mode.
 **Use case 2:** Login in demo mode, logout, and login using normal login

### :art: UI changes?


xurxodev (approved):
thanks @idelcano

PR #2193, opened 2018-11-13
Create a controller to init dhis in the EyeSeeTeaApplication.

### LINKED ISSUE #2191: Remove dependencies from dhis SDK in EyeSeeTeaApplication
On EyeSeeTeaApplication we do this:
D2.init(this);
If we use other backend we can not have this depdendency here. Please analize where we can to have this D2.init

manuelplazaspalacio:
### :pushpin: References
* **Issue:** close #2191 

Remove dhsi2 dependencies from EyeSeeTeaApplication

Calling D2 init when login or when pushing.

Run the app an login an check that pull and push works correctly.

### :art: UI changes?


xurxodev (changes requested):
@manuelplazaspalacio With this solution, you are assuming that either dhis 2 sdk should be initialize passing EyeSeeTeaApplication but for example for new SDK in kotlin this is not necessary.
We should not assume that from EyeSeeTeaApplication we initilize a dhis sdk, this code should be remove from EyeSeeTeaApplication. 
If an dhis2 SDK must be initialize is in adapters where this should be realized.
Please review this solution.

xurxodev (approved):
thanks @manuelplazaspalacio

PR #2231, opened 2019-01-17
Feature fix api calls with missing slash

### LINKED ISSUE #2228: Add / at the end of the URL only when needed

### LINKED ISSUE #2229: quarantine check URL is failing

idelcano:
### :pushpin: References
* **Issue:** close #2228 and close #2229 

###   :gear: branches 
**app**: 
       Origin: feature-fix_api_calls_with_missing_slash Target: v1.3_hnqis
**bugshaker-android**: 
**dhis2-light-sdk**: 
Origin: development
**EyeSeeTea-SDK**: 
Origin: development
**SDK**: 
Origin: 2.25_EyeSeeTea

Add / at the end of the URL only when needed
quarantine check URL is failing 

Note: I did the two issues at the same time because when I refactorized executecall methods to make it simple the quarantine error was solved.

The problem with quarantine check was in an error adding the data to the server in the executecall methods. (url+data+data). I removed that error refactoring to complete the other issue.

- I have added a slash on the URL in the server string when the user doesn't include the final "/"

- I have removed all the API methods switch and param because we are only using the "get" method.

- I have removed all the starter slash in the API methods.

** Use case 1: ** - Realize login with https://data.psi-mis.org and you should login ok

 ** Use case 2: ** - Realize login with https://data.psi-mis.org/ and you should login ok

 ** Use case 3: ** - Do push and check the checkQuarantine call on the log is success.
Note: In v1.3_hnqis if a push fails the next push is in 30 mins... You can change it in
private static final long PUSH_FAIL_PERIOD = 300L;
To provoke the checkQuarantine API call do a push with a breakpoint in PushController line 170. When the debugger move to the breakpoint close the app, and reopen.
Check the log to see the events API call.
You will show some similar than:

 ```
executeCall Url https://data.psi-mis.org/api/events.json?program=mZ0zJcGODwD&orgUnit=A5HZNrJc2ir&startDate=2019-01-17&endDate=2019-01-25&skipPaging=true&fields=event,orgUnit,program,dataValues

D/.CheckSurveysB&D: Set quarantine survey as complete 818 date 2019-01-17T11:31:04.011+0000

D/org.eyeseetea.malariacare.data.database.model.SurveyDBB&D: Id: 818 actual status:1 set as:1
```

### :art: UI changes?

### DIFF
--- a/app/src/main/java/org/eyeseetea/malariacare/data/remote/api/OkHttpClientDataSource.java
+++ b/app/src/main/java/org/eyeseetea/malariacare/data/remote/api/OkHttpClientDataSource.java
@@ -26,13 +26,23 @@ public class OkHttpClientDataSource {
 
     /**
      * Call to DHIS Server
-     * @param data
-     * @param method
+     * @param apiCall
+     */
+    public static Response executeCall(String apiCall) throws IOException {
+        return executeCall(new BasicAuthenticator(PreferencesState.getInstance().getCreedentials()), PreferencesState.getInstance().getServer().getUrl(), apiCall);
+    }
+
+    /**
+     * Call to DHIS Server
      * @param url
      */
-    public static Response executeCall(BasicAuthenticator basicAuthenticator, JSONObject data, String url, String method) throws IOException {
-
-        Log.d(TAG, "executeCall Url" + url + "");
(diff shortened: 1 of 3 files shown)


xurxodev (approved):
thanks @idelcano

PR #2298, opened 2019-05-27
Feature/show message to user if required metadata in observations not exist

### LINKED ISSUE #2297: When a required data element for observations does not exist in the database then should show a message to the user

xurxodev:
### :pushpin: References
* **Issue:** close #2297   
* **Related pull-requests:**

###   :gear: branches 
**app**: 
       Origin: feature/show_message_to_user_if_required_metadata_in_observations_not_exist Target: v1.5_hnqis
**bugshaker-android**: 
**EyeSeeTea-SDK**: 
       Origin: development   
**SDK**: 

Avoid a crash in observations when required data elements do no exists

I have created validations in the repository and throw an exception if does not exists some required data element
I have modified ObservationsPresenter and ObservationFragment to catch the new exception and show the message to the user and disable the observation screen.

 **Use case 1:** - Login and pull with all new control data elements added to a program, The observations screen should work successfully

 **Use case 2:** - Login and pull without new control data elements added to a program, The observations screen should show the message 'An error has occurred, there is no metadata necessary to create action plan' and the controls should be disabled.

### :art: UI changes?


ifoche (approved):
thanks @xurxodev

PR #2304, opened 2019-05-29
Bug/filters sent list missing when back from action plan

### LINKED ISSUE #2303: Bug: When navigating to action plan directly without open feedback screen and click on the back button, then the improve list has missed the orgUnit and program filters

xurxodev:
### :pushpin: References
* **Issue:** close #2303 
* **Related pull-requests:**

###   :gear: branches 
**app**: 
       Origin: bug/filters_sent_list_missing_when_back_from_action_plan Target: v1.5_hnqis
**bugshaker-android**: 
**EyeSeeTea-SDK**: 
       Origin: development   
**SDK**: 

There is a bug in HNQIS 1.4 version and We are going to fix it to the 1.5 version:

Open the app the first time and navigate to any action plan directly (not navigate to feedback previously) then click on back button navigating to sent list again but the Orgunit and program filters on the top of the tab are missing.

This bug does not occur if previously you had navigated to feedback because back from feedback to sent list work and if the feedback is visited previously, from action plan to click on back navigate to feedback and does not navigate to sent list.

**It will take quite some time to discover the error because the part of tabs and controllers is very confusing**
The problem was that the filter view is hidden before to navigate to the action plan and It's necessary to mark it as visible when back to the sent list.

 **Use case 1:** - Open the app the first time and navigate to any action plan directly (not navigate to feedback previously) then click on back button navigating to sent list again and the Orgunit and program filters on the top of the tab show be visible.

### :art: UI changes?


ifoche (approved):
thanks @xurxodev

PR #2362, opened 2019-11-12
Avoid bugs caching all surveys to create ObservationLocalDataSource

### LINKED ISSUE #2351: Unresponsive Action Plan
At times all the fields in Action plan are unresponsive and it is not possible to type anything in. The failed steps do not populate either. We had to uninstall and reinstall the app several times to finally get it to work. At times it is working then if you exit the action plan before you complete typing, when you get back into the action plan, all the failed steps that had populated vanish, then the app bounces to Plan module. See
https://drive.google.com/drive/folders/1UqRR3CsSNk84m0s_hUNbN932OYXoJh5f

xurxodev:
### :pushpin: References
* **Issue:** close #2351   
* **Related pull-requests:**

###   :gear: branches 
**app**: 
       Origin: maintenance/change_strings_requested_by_client Target: v1.5_hnqis
**bugshaker-android**: 
**EyeSeeTea-SDK**: 
       Origin: development 
**SDK**: 

At times all the fields in Action plan are unresponsive and it is not possible to type anything in. The failed steps do not populate either. We had to uninstall and reinstall the app several times to finally get it to work. At times it is working then if you exit the action plan before you complete typing, when you get back into the action plan, all the failed steps that had populated vanish, then the app bounces to Plan module. See
https://drive.google.com/drive/folders/1UqRR3CsSNk84m0s_hUNbN932OYXoJh5f

After many many tests I only have reproduced this behaviour in the next scenario:
1 - Realize login and pull
2- Realize Logout
3 - Realize login and pull
4 - Start a new observation and exist from observation changing of tab for example
5 - Navigate to the same observations

The bug is because DataFactory (object in kotlin = singleton in java) only creates one instance of ObservationLocalDataSource and this one realizes a cache query of all surveys. then after logout, this cache is not cleared.
This cache of surveys can provoke other errors.

I have removed this cache.
I have modified ObservationLocalDataSource to create mapper and query necessary survey o surveys on every method that uses the mapper.

**Use case 1:** - realize the scenario
1 - Realize login and pull
2- Realize Logout
3 - Realize login and pull
4 - Start a new observation and exist from observation changing of tab for example
5 - Navigate to the same observations
Then the observation data should be rendered.

### :art: UI changes?

### DIFF
--- a/app/src/main/java/org/eyeseetea/malariacare/data/database/datasources/ObservationLocalDataSource.java
+++ b/app/src/main/java/org/eyeseetea/malariacare/data/database/datasources/ObservationLocalDataSource.java
@@ -2,7 +2,6 @@ package org.eyeseetea.malariacare.data.database.datasources;
 
 import com.raizlabs.android.dbflow.sql.language.Delete;
 import com.raizlabs.android.dbflow.sql.language.From;
-import com.raizlabs.android.dbflow.sql.language.SQLCondition;
 import com.raizlabs.android.dbflow.sql.language.Select;
 import com.raizlabs.android.dbflow.sql.language.Where;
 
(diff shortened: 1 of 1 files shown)


ifoche (approved):
thanks @xurxodev

PR #2374, opened 2019-11-18
Change changeSelectedFilters in OrgUnitProgramFilterPresenter to safe…

### LINKED ISSUE #2350: Random Crashing Behavior
1.5.1 is crashing quite a bit. This is random, it could be just after loading the app and you want to log in, or within the app when moving from one module to another or when you want to start conducting an assessment after selecting the HA and Org Unit. We have experienced this on 4 different devices and different servers. The other versions of the app, 1.4.2 and 1.4.6 are not exhibiting this behavior while testing under the same conditions.

xurxodev:
### :pushpin: References
* **Issue:** close #2350  
* **Related pull-requests:**

###   :gear: branches 
**app**: 
       Origin: maintenance/improve_filters_presenters_code_from_bugs Target: v1.4_hnqis
**bugshaker-android**: 
**EyeSeeTea-SDK**: 
       Origin: development 
**SDK**: 

The client reports a random crashing and I have not replicate the crash.
But after review Crashlytics I have viewed this bug:
https://fabric.io/eyeseetea/android/apps/org.eyeseetea.malariacare.hnqis_ng_dev/issues/2c133f2aa5411994844d027507f1cc63

I think is possible after logout and login with another user with different assigned programs or OrgUnits the changeSelectedFilters crash in the next scenario:

User change program or org unit filter and these uids are stored in share preferences then logout and login with another user with other assigned programs or org units. In this case, changeSelectedFilters maybe crash because the first Kotlin method crash if the element does not exist.

I goal is refactor changeSelectedFilters to safer code.

- I have refactored changeSelectedFilters to use firstOrNull method and Elvis operator

**Use case 1:** - Filters should work on every tab

### :art: UI changes?


ifoche (approved):
thanks @xurxodev

PR #2422, opened 2020-08-04
Feature/reading server mode from poeditor

### LINKED ISSUE #2420: Parse POEditor term

xurxodev:
### :pushpin: References
* **Issue:** close #2420     
* **Related pull-requests:**

###   :gear: branches 
**app**: 
       Origin: feature/reading_server_mode_from_poeditor Target: v1.6_hnqis
**bugshaker-android**: 
**EyeSeeTea-SDK**: 
       Origin: development 
**SDK**: 

Now the servers retrieved from PoEditor should have the assigned classification (COMPETENCIES OR SCORING).
The goal is to parse this info and save in the database

- I have created a new Enum ServerClassficiation
- I have created a new property classification in the Server entity 
- I have parsed this info in the ServerRemoteDataSource and create integration tests using MockWebServer to validate the behaviour
- I have created a new classification column in server table and It's included during mappings in reading and saving processes.
- I have created a migration to add the classification column

to test different server responses until po editor return classification info by server, this code help to this goal:

In PoEditorApiClient class insert this code replacing line 28:
```
val hackedTerm = Term(
    "server_list", Date(), Translation(
        "https://data.psi-mis.org/|SCORING\n" +
            "https://ao-dev.hnqis.org/|Scoring\n" +
            "https://imdatahub.org/|scoring\n" +
            "https://myanmar.psi-mis.org/|SCORING\n" +
            "https://sfhza.psi-mis.org/|SCORING\n" +
            "https://sm-dev.hnqis.org/|SCORING\n" +
            "https://zw.hnqis.org/|SCORING"
    )
)

val mutableTerms = terms.toMutableList()

mutableTerms[247] = hackedTerm
Either.Right(mutableTerms)
```

**use case 1**: Start the app with a login realized from las version and the migration should be executed and assigned competencies (code 1) classification by default to all servers in the database
**use case 2**: Realize log out, close the app, insert the code to simulate a server response in line 128 of PoEditorApiClient, then execute the app and the servers in the database should contain the classification according to simulated server response
**use case 3**: Remove the app and start from new installation and the servers in the database should contain the classification according to simulated server response

- [X] Yes, it's already applied.

### :art: UI changes?


ifoche (approved):
thanks @xurxodev . I'm approving the PR, but I found a minor thing that we can deal with later on. If you configure in POEditor "key|" and don't introduce anything after, the error produces an infinite spinning wheel on the app. I think we should either display an error and assume default or just assume default type, but avoid crashing

PR #2439, opened 2020-09-18
Fix crash after refresh metadata and fix crash after change language

### LINKED ISSUE #2438: 1.6.0 Crash when refreshing metadata
Description
--
App stops when finishing the 'refresh metadata' process
server:https://data.psi-mis.org/User:KEdemo1

How to replicate
--
Click on the 3 dots menu of the right top of the screenSelect Refresh metadata

xurxodev:
### :pushpin: References
* **Issue:** close #2438 and #2440 
* **Related pull-requests:**

###   :gear: branches 
**app**: 
       Origin: fix/crash_after_refresh_metadata Target: v1.6_hnqis
**bugshaker-android**: 
**EyeSeeTea-SDK**: 
       Origin: development 
**SDK**: 

Fix crash after refresh metadata and fix crash after change language

After many tests, I have found the problem. This bug seems related to commit fragments from the OnResume method in the activity.

I have followed the advice in https://www.androiddesignpatterns.com/2013/08/fragment-transaction-commit-state-loss.html

First I have removed the reloadActiveTab invocation from onResume in the DashboardActivity that replaces the current fragment using commit. 

I think this is not a problem because after back from inProgress o settings activity to DashboardActivity the current tab has been created in the onCreate method from the activity and when the current tab change we reload the tab.

However, this provokes other errors in or app with our special approach to manage fragments and tabs.

So I have used the commitAllowingStateLoss() only as a last resort as recommended the article

**use case 1**: Log in and pull should work and navigate to the dashboard without crash
**use case 2**: Refresh metadata should work and navigate to the dashboard without crash
**use case 3**: Navigate to settings to change language and back to the Dashboard Activity should work without any crash.

### :art: UI changes?

### DIFF
--- a/app/src/main/java/org/eyeseetea/malariacare/BaseActivity.java
+++ b/app/src/main/java/org/eyeseetea/malariacare/BaseActivity.java
@@ -387,9 +387,8 @@ public abstract class BaseActivity extends AppCompatActivity {
     }
 
     protected void goSettings() {
-        Intent intentSettings = new Intent(this, SettingsActivity.class);
-        intentSettings.putExtra(SETTINGS_CALLER_ACTIVITY, this.getClass());
-        startActivity(new Intent(this, SettingsActivity.class));
+        Intent intent = new Intent(this, SettingsActivity.class);
+        ActivityCompat.startActivity(this, intent, null);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


ifoche (approved):
thanks @xurxodev

PR #2452, opened 2020-11-11
Create new server in db if logged server does not exist

### LINKED ISSUE #2451: Bug during the login and pull if the logged server is inserted manually (https://clone.psi-mis.og)

xurxodev:
### :pushpin: References
* **Issue:** close #2451  
* **Related pull-requests:**

###   :gear: branches 
**app**: 
       Origin: maintenance/fix_bug_in_pull_if_insert_server_manually Target: v1.6_hnqis
**bugshaker-android**: 
**EyeSeeTea-SDK**: 
       Origin: development 
**SDK**: 

When the user inserts the server manually, this server is not created in the server table. This provokes an error during the login/pull .

- When in the LoginUseCase after login, we are going to update the server in the database as conected, if the server does not exist then we create it previously.

**use case 1**: Realize login against a server that it's not in the list, after login the server should be created in the database and mark as connected.

### :art: UI changes?

### DIFF
--- a/app/src/main/java/org/eyeseetea/malariacare/domain/usecase/LoginUseCase.java
+++ b/app/src/main/java/org/eyeseetea/malariacare/domain/usecase/LoginUseCase.java
@@ -37,7 +37,7 @@ import java.net.MalformedURLException;
 import java.net.UnknownHostException;
 import java.util.List;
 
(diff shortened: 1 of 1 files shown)


ifoche (approved):
thanks @xurxodev

PR #2457, opened 2021-01-11
Fix bug with authority changing the server

xurxodev:
### :pushpin: References
* **Issue:** close https://app.clickup.com/t/caxf1t   
* **Related pull-requests:**

###   :gear: branches 
**app**: 
       Origin: fix/bug_with_authority_changing_of_server Target: v1.6_hnqis
**bugshaker-android**: 
**EyeSeeTea-SDK**: 
       Origin: development 
**SDK**: 

the problem was that the api service (Retrofit) was not recreated when the url change, the solution is to recreate retrofit and api service when the server url and last server url are different

- I have modified UserD2ApiRepository to recreate api service when is necessary

#### Case 1
1. Try to login with the following credentials: server (https://clone.psi-mis.org), user(KEdemo1), password (Kenyademo1!)
You should not be able to log in.
2. Change server to https://data.psi-mis.org
3. You should be able to log in.
#### Case 2
1. Login with the following credentials: server (https://data.psi-mis.org), user(KEdemo1), password (Kenyademo1!)
2. You should be able to log in
3. Log out
4. Login with the following credentials: server (https://clone.psi-mis.org), user(KEdemo1), password (Kenyademo1!)
5. You should not be able to log in

### :art: UI changes?

### DIFF
--- a/app/src/main/java/org/eyeseetea/malariacare/data/repositories/UserD2ApiRepository.kt
+++ b/app/src/main/java/org/eyeseetea/malariacare/data/repositories/UserD2ApiRepository.kt
@@ -13,7 +13,8 @@ import retrofit2.converter.gson.GsonConverterFactory
 
 class UserD2ApiRepository : UserRepository {
 
-    private var userD2Api: UserD2Api? = null
+    private lateinit var userD2Api: UserD2Api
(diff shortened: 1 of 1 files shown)


ifoche (approved):
thanks @xurxodev

PR #2463, opened 2021-06-11
Send server as user property to analytics

xurxodev:
### :pushpin: References
* **Issue:** https://app.clickup.com/t/m3bbc6
* **Related pull-requests:**

###   :gear: branches 
**app**: 
       Origin: feature/firebase_analytics_send_user_properties Target: v1.6_hnqis
**bugshaker-android**: 
       Origin: development_android_x  
**EyeSeeTea-SDK**: 
       Origin: feature/development 
**SDK**: 

Filter reports in firebase by the server (crashlytics and analytics)

- [x] Send user property server to analytics
- [x] Add android.permission.WAKE_LOCK, this is necessary to send user properties by analytics instance

I have investigated crashlytics to filter by sent keys from the app.  We have several options:

- from dashboard page we can filter by bugs with specific key name

Captura de pantalla 2021-06-10 a las 17 16 21

- from the bug page, we can filter the events of a bug using key-value

Captura de pantalla 2021-06-10 a las 17 09 43

I have investigated analytics to filter by sent keys in crashlytics and this is not possible. We need to send new info from the app for analytics.
I think the best option for us is to send a new server user property, then we can filter in the dashboard and for events using server user property.
Captura de pantalla 2021-06-11 a las 11 58 22

Captura de pantalla 2021-10-11 a las 9 23 40

**important**: in the firebase console is necessary to create a new custom dimension from the custom definitions menu in the sidebar. The new dimension must be to have as user property server. I have created it in debug firebase but we need to create it in production firebase.

**Use case 1**:  should send server user property to the analytics

### :art: UI changes?

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -13,6 +13,7 @@
         tools:node="remove" />
     <uses-permission
         android:name="android.permission.READ_EXTERNAL_STORAGE" />
+    <uses-permission android:name="android.permission.WAKE_LOCK" />
     <uses-permission
         android:name="android.permission.WRITE_EXTERNAL_STORAGE"/>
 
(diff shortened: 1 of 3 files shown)


ifoche (approved):
thanks @xurxodev

PR #2468, opened 2022-02-07
Fix bug exporting database in devices from api 29

xurxodev:
### :pushpin: References
* **Issue:** 
* **Related pull-requests:**

###   :gear: branches 
**app**: 
       Origin: fix/export_database Target: v1.6_hnqis
**bugshaker-android**: 
       Origin: development_android_x  
**EyeSeeTea-SDK**: 
       Origin: feature/development 
**SDK**: 

Fix export database

After testing in some devices, physical and emulator. the problem only occurs with devices with API greater than 28.
The problem is to create the txt file with metadata where we try insert IMEI data for example.
This data is not possible to retrieve from the device from Android 10.
https://source.android.com/devices/tech/config/device-identifiers

- [x] Fix export database from Android 10

- [ ] Export database should work in old and new devices

### :art: UI changes?


ifoche (approved):
thanks @xurxodev

PR #2469, opened 2022-05-26
Remove lateinit for programs and orgUnits

xurxodev:
### :pushpin: References
* **Issue:**  https://app.clickup.com/t/2c66ncc

###   :gear: branches 
**app**: 
       Origin: feature/fix_late_init_bugTarget: v1.6_hnqis
**bugshaker-android**: 
       Origin: development_android_x  
**EyeSeeTea-SDK**: 
       Origin: feature/development 
**SDK**: 

Fix kotlin.UninitializedPropertyAccessException: lateinit property programs has not been initialized

I have not reproduced the error, but programs are loaded by use case then I think it's possible if the response is low that other methods where the program is used be invoked before programs are set. I have removed lateinit keyword and set it to an empty list by default.

- [x] remove lateinit from programs and org unit and set to empty list by default

### :art: UI changes?


ifoche (approved):
thanks @xurxodev

PR #2471, opened 2022-05-26
Fix crash executing push after logout

xurxodev:
Fix null pointer exception to delete survey:
- When not exists last completion survey
- ### :pushpin: References
* **Issue:**  https://app.clickup.com/t/2c66nr9

###   :gear: branches 
**app**: 
       Origin: fix/crash_executing_push_from_logoutTarget: v1.6_hnqis
**bugshaker-android**: 
       Origin: development_android_x  
**EyeSeeTea-SDK**: 
       Origin: feature/development 
**SDK**: 

Fix crash when push is executing during logout process and before cancel alarm

If load credentials is executed and shared preferences for url, user, and password are empty then set credentials to null in session. When from push service the credentials are retrieved from the session if credentials are null then not execute the push

- [x] Fix crash when push is executing during logout process and before cancel alarm 

It's difficult to reproduce

- Execute logout at the same time the push is executed
- if the logout is executed first, the push should be cancel by alarm or null credentials

### :art: UI changes?


ifoche (approved):
thanks @xurxodev

PR #2472, opened 2022-05-26
Fix Parameter specified as non-null is null: crash

xurxodev:
Fix null pointer exception to delete survey:
- When not exists last completion survey
- ### :pushpin: References
* **Issue:** 
 https://app.clickup.com/t/2c66nmm 
https://app.clickup.com/t/2c66p7y

###   :gear: branches 
**app**: 
       Origin: fix/crash_parameter_specified_as_non_null_is_null Target: v1.6_hnqis
**bugshaker-android**: 
       Origin: development_android_x  
**EyeSeeTea-SDK**: 
       Origin: feature/development 
**SDK**: 

Fix crash `Fatal Exception: java.lang.NullPointerException: Parameter specified as non-null is null: method `

I have not reproduced the bug, but I have applied the solution that I have found here:
https://stackoverflow.com/a/45255651

- [x] Fix Parameter specified as non-null is null

### :art: UI changes?


ifoche (approved):
thanks @xurxodev

PR #2477, opened 2022-06-23
Close cursor after to use it in migrations

xurxodev:
- ### :pushpin: References
* **Issue:**  https://app.clickup.com/t/2c66p2w

###   :gear: branches 
**app**: 
       Origin: fix/database_cursor_window_allocation_exception Target: v1.6_hnqis
**bugshaker-android**: 
       Origin: development_android_x  
**EyeSeeTea-SDK**: 
       Origin: feature/development 
**SDK**: 

Fix data base CursorWindowAllocationException

This bug is complicated, It can be provoked by dbFlow, I have found some issues in the library GitHub
https://github.com/agrosner/DBFlow/issues/1642
https://github.com/agrosner/DBFlow/issues/1700

In the code where the bug is thrown, we don’t use a cursor directly, only we use dbflow models to query the database

```
public static ServerMetadataDB findControlDataElementByCode(String code){
    return new Select().from(ServerMetadataDB.class).where(ServerMetadataDB_Table.code.eq(code)).querySingle();
}
```

I have investigated this error and the problem seems related to memory usage and open database cursors.

We only use cursors directly in migrations. I don’t know if the problem is our cursors or internal dbflow cursors. I have reviewed our cursors in the migrations and someone is not closed after use.

We can try to close cursors in migrations where it’s not closed and review in the future if this bug appears again.

- [x] Close cursor after to use it in migrations

### :art: UI changes?

### DIFF
--- a/app/src/main/java/org/eyeseetea/malariacare/data/database/migrations/Migration17AddObservationValueTable.java
+++ b/app/src/main/java/org/eyeseetea/malariacare/data/database/migrations/Migration17AddObservationValueTable.java
@@ -26,86 +26,106 @@ public class Migration17AddObservationValueTable extends BaseMigration {
 
     @Override
     public void migrate(DatabaseWrapper database) {
-        Cursor obsActionPlanDBCursor = database.rawQuery(
-                "Select id_obs_action_plan, id_survey_obs_action_fk, provider, gaps, "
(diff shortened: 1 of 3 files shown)


ifoche (approved):
thanks @xurxodev

PR #2493, opened 2022-12-26
Refactor fragment and services

xurxodev:
### :pushpin: References
* **Issue:** https://app.clickup.com/t/33e685x
* related PR * https://github.com/EyeSeeTea/dhis2-android-sdk/pull/437

###   :gear: branches 
**app**: 
       Origin: feature/services Target: v1.6_hnqis
**bugshaker-android**: 
       Origin: development_android_x  
**EyeSeeTea-SDK**: 
       Origin: feature/development 
**SDK**: 
       Origin: feature/upgrade_target_sdk_31

Testing these fixed errors is complicated because the errors were produced in background. The error was thrown when a service started with the app in background.

The more simple way to test is review all fragments and review reloads is working. Then review firebase crashlytics and errors similar to this should not occur:

```
Caused by: java.lang.IllegalStateException: Not allowed to start service Intent { cmp=my.app.tt/com.my.service }: app is in background uid UidRecord{90372b1 u0a136 CEM idle procs:1 seq(0,0,0)}

```
```
The startService() method now throws an IllegalStateException if an app targeting Android 8.0 tries to use that method in a situation when it isn't permitted to create background services.

```
The most simple 

### :art: UI changes?


ifoche (approved):
thanks @xurxodev

PR #2495, opened 2023-05-26
Hide button but maintaining the space.

xurxodev:
### :pushpin: References
* **Issue:** https://app.clickup.com/t/860qy9f7p
* related PR * 

###   :gear: branches 
**app**: 
       Origin: fix/overlapsed_text_in_assessment_when_no_survey_is_found  Target: v1.6_hnqis
**bugshaker-android**: 
       Origin: development_android_x  
**EyeSeeTea-SDK**: 
       Origin: feature/development 
**SDK**: 

Fix overlapped text

- [x] - Fix overlapped text

- From Android P you cannot use setVisibility directly. After investigating, the recommended approach is to use show o hide methods but there is no equivalent for set visibility to invisible VisibilityAwareImageButton.setVisibility can only be called from within the same library group (referenced groupId=com.google.android.material from groupId=malariapp)

I use a hack to convert the fab to view

### :art: UI changes?


ifoche (approved):
thanks @xurxodev

